PR #6, opened 2017-10-17
updated main.cpp

plex1:
added missing using namespace std

i couldn't build the repo without this change


Dolu1990:
Hooo right. Thanks !
I don't know why, but on my setup it doesn't catch it, and is also happy without it XD

PR #8, opened 2017-10-27
Fixed an issue with bursted AXI Transactions on the ICache

drichmond:
Size must be set to the number of bytes in a single burst cycle (beat). At the
moment it is set to 0 (unused) in InstructionCache.scala. As a consequence,
when a burst request is made by the cache the slave returns a 32-bit value
repeatedly because the slave thinks that it has passed the wrap boundary.

This fix implements logic for size in the InstructionCache

### DIFF
--- a/src/main/scala/vexriscv/ip/InstructionCache.scala
+++ b/src/main/scala/vexriscv/ip/InstructionCache.scala
@@ -29,7 +29,7 @@ case class InstructionCacheConfig( cacheSize : Int,
     useRegion = false,
     useLock = false,
     useQos = false,
-    useSize = false
+    useSize = true
   )
 
   def getAvalonConfig() = AvalonMMConfig.bursted(
@@ -134,6 +134,7 @@ case class InstructionCacheMemBus(p : InstructionCacheConfig) extends Bundle wit
     mm.readCmd.addr := cmd.address
     mm.readCmd.prot  := "110"
     mm.readCmd.cache := "1111"
+    mm.readCmd.size := log2Up(p.memDataWidth/8)
     if(p.wrappedMemAccess)
       mm.readCmd.setBurstWRAP()
     else


Dolu1990:
Hi,
In the AXI4 specification, Table A10-3 Master interface read channel signals and default signals values,  
Signal name | Default
ARSIZE         | Data bus width

So if the signals is not present, it is implicitly 4 bytes transactions on a 32 bits bus, right ?
(it was intentional)

drichmond:
That seems to be the case...

If you connect a VexRiscV to a xilinx AXI interconnect it seems to default to zero though - which causes the behavior 

Your call

Dolu1990:
When you connect the bus to their IP, could you wire those additional wire ? or you can't because it is a whole bus connection with a single wire (graphical interface)

Then one  solution could be to add an parameter in the VexRiscv generation to ask it to add those signals which aren't handeled properly by some other tools ?

(I don't realy like doing thing wrong because other tools are wrong :))

drichmond:
I used IPI - so a graphical interface.

Dolu1990:
Ok, then, for instance, if the CPU is your toplevel, you can do this kind of things :

```scala
     SpinalVerilog{
      val cpu = VexRIscv(.....)
      cpu.rework {
        val arsize = out(B"010").setName("io_dbus_axi_miaou_arsize")
      }
     cpu
   }
```

It will add the an output wire to your cpu which will be drived with the value 2.
Would it will be ok ? Then Xilinx tool will be happy, without having to modify the CPU implementation

drichmond:
Yeah - if that works I’m happy with it. I’ll check when I get back on Monday

Since we’re on the topic, is there a way to change the signal naming? Xilinx doesn’t like the _payload and other underscores when automatically identifying busses (and this may be a reason the tools are doing the wrong thing )

Dolu1990:
Sure, about renaming things, you can for instance use the Axi4SpecRenamer tool. You can give as argument of Axi4SpecRenamer a AXI bus, then signals of it will be AXI like named

I don't realy know how your toplevel look like, but for instance : 
master(Axi4SpecRenamer(plugin.dBus.toAxi4Shared()))

imported via 
import spinal.lib.bus.amba4.axi.Axi4SpecRenamer

drichmond:
Bringing this back - Where did you see the default AXI behavior?

Every implementation I've looked at in the past few weeks sets the ARSIZE value and never leaves it unconnected. Talking to folks at Xilinx echo the same feeling - while it may say that in the AXI spec you shouldn't necessarily rely on the default.

Dolu1990:
The default AXI behavior can be found on the ARM IHI 0022D pdf, at "Table A10-3 Master interface read channel signals and default signals values"
There ARSIZE is taged as an optional signal, with a "Data bus width" default value.

"Every implementation I've looked at in the past few weeks sets the ARSIZE value and never leaves it unconnected"  I think it is mostly due to limitation of VHDL/Verilog/EDA flow which endup with those (xilinx) practices.

Here the concept isn't to keep things unconnected, but voiding their existance. Defining an AXI interface doesn't mean having to define all wires if for some of them the default value match your usage.  (is it wrong ?)  (In SpinalHDL you can can define interfaces where some pins existance is parametrized)

Also SpinalHDL when you connect to AXI busses together the connection operator  take care of applying default values to the slave side if the master doesn't provide the corresponding wire. Also if a slave doesn't provide a pin that the master provide, it will tell you that something is wrong.

So just to be sure if we are thinking the same (tell me if i'm wrong): 
- Xilinx is wrong (Maybe because of legacy / EDA / HDL stuff)
- But as they are a big player, the true AXI specification (in this FPGA area) is the xilinx interconnect implementation (kind of)

Dolu1990:
Hooo if found this in the AXI spec : 

> A10.3 Default signal values
> This specification suggests that, in general, for maximum IP reuse, an AXI component interface includes all signals.
> The presence of all signals reduces the risk of error at the system integration phase of the design flow and it can also help support some design flows that do not effectively support default values for absent signals.

What i can do, is to add a setting to ask VexRiscv  to keep all AXI signals for an easy integration :)

drichmond:
I think that's a fair compromise

Dolu1990:
Ok, so, i made an update.
Basicaly it use some an unreleased version of SpinalHDL, you will need to do :
```
git clone https://github.com/SpinalHDL/SpinalHDL.git
cd SpinalHDL
sbt clean publish-local
```

in order to get this unreleased version on you PC.

Then you can get the VexRiscv dev branch, there you will find a Axi demo :
https://github.com/SpinalHDL/VexRiscv/blob/ce6fd6d0aa88e7f2e2369f75e74b013537f57e5b/src/main/scala/vexriscv/demo/VexRiscvAxi4WithIntegratedJtag.scala#L156

This line :
master(plugin.dBus.toAxi4Shared().toAxi4().toFullConfig())

It's taking the CPU native data bus (dBus), then bridge it to Axi4Shared (an unoffical Axi4 variation), then bridge it into the standard Axi4, and finaly bridge it into a full Axi4 for a better integration.

It is many bridges, but in fact it is nearly logic less ones.

<3 SpinalHDL <3

Dolu1990:
Oups, missed the toFullConfig for the iBus.
there is the corrected one : 
https://github.com/SpinalHDL/VexRiscv/blob/9b9bbaa4ad1c7c3ce0cb20a7fd2461270bd043c6/src/main/scala/vexriscv/demo/VexRiscvAxi4WithIntegratedJtag.scala#L156

Dolu1990:
Can i close this issue ?

drichmond:
Yes

Dolu1990:
Ok,
Thanks for the feeds back :)

PR #9, opened 2017-11-16
Minor fixes to Murax.scala

kaofishy:
Changes GPIO width from hard coded to gpioWidth so it can be adjusted through instantiation parameter.
Adds explicit type to apbDecoder.slave to suppress IDE errors in IntelliJ.


Dolu1990:
Great thanks !

PR #24, opened 2018-06-04
Fix some missing Barriel -> barriel fixes

tomverbeure:
My earlier fix broke travis.

Tom


Dolu1990:
Hmm i haven't gived attention to travis on the master branch, the real checks are currently only in the fetchRework branch, will be merged into master soon

PR #27, opened 2018-06-19
README language

tomverbeure:


Dolu1990:
Thanks ^^

Just, why is there some line return at (for instance) https://github.com/SpinalHDL/VexRiscv/pull/27/files#diff-04c6e90faac2675aa89e2176d2eec7d8R850 ?

tomverbeure:
I simply break up long lines to a reasonable length.

This is personal preference I guess for those who don't have auto-wrap enabled in their editor.

Tom

Dolu1990:
ok ^^
Thanks

Issue #48, opened 2018-12-22
JTAG/Reset Lock Up (Avalon)


cutephoton:
So I've been banging my head on the desk trying to understand the debug/jtag. I think the problem is related to reset.

For reference, this is my qsys diagram:

![qsys](https://user-images.githubusercontent.com/19512121/50374047-d2222180-059c-11e9-8ee9-69b3ebd80dcf.png)

`monitor reset halt` -> locks up
`monitor reset resume` -> locks up
`monitor halt` -> works
`monitor resume` -> works

This is what halt/resume cycles look like:

![image](https://user-images.githubusercontent.com/19512121/50375038-2c76ae80-05ac-11e9-9c0e-0edc34b2a147.png)

And now `monitor reset run/halt`

![image](https://user-images.githubusercontent.com/19512121/50375060-795a8500-05ac-11e9-9fde-ccd3ad5b818e.png)

Thoughts?

OCD config:
```
#-----------------------------------------
# JTAG Adapter
#
interface ftdi
ftdi_vid_pid 0x0403 0x6014
#ftdi_layout_init 0x0c08 0x0f1b
ftdi_layout_init 0xfff8 0xfffb
ftdi_layout_signal nTRST -data 0x0100 -oe 0x0100
ftdi_layout_signal nSRST -data 0x0200 -oe 0x0200
transport select jtag

#-----------------------------------------
# Target
#
set  _ENDIAN little
set _TAP_TYPE 1234
set _CPUTAPID 0x10001FFF

adapter_khz 4000
# have tried 25,100,1000 too
adapter_nsrst_delay 260
jtag_ntrst_delay 250

set _CHIPNAME fpga_spinal
jtag newtap $_CHIPNAME bridge -expected-id $_CPUTAPID -irlen 4 -ircapture 0x1 -irmask 0xF

target create $_CHIPNAME.cpu0 vexriscv -endian $_ENDIAN -chain-position $_CHIPNAME.bridge -coreid 0 -dbgbase 0xF00F0000
vexriscv readWaitCycles 12
vexriscv cpuConfigFile ../cpu0.yaml

poll_period 50

init
#echo "Halting processor"
soft_reset_halt
```
GDB:
```
target remote localhost:3333
set arch riscv:rv32
monitor reset halt
load
continue
```

Dolu1990:
Hi,
[...]

PR #49, opened 2018-12-22
Avalon: Debug Clock Domain for JTAG

cutephoton:
This change ensures that the clock domain for the JTAG interface
uses the debug plugin's domain. Otherwise, resetting the processor
will put the jtag debugger in to reset as well.

See SpinalHDL/VexRiscv#48


cutephoton:
just a thought to take this one step further:
Is there a way to put an assertion or warning to check the clock domains are different?

I think it might be helpful to find a way to put some warnings in to help identify potential issues.

PR #52, opened 2019-01-19
[WIP] Docker images

eine:
Related to SpinalHDL/SpinalHDL#195.

In this PR, using docker image `spinalhdl/dev` is proposed as an alternative to installing the dependencies in each Travis test/job.

While testing the current implementation, I found it surprising that tests pass when using `spinalhdl/dev`, where SpinalHDL is not installed: https://travis-ci.com/1138-4EB/VexRiscv/builds/97934350. I would expect SpinalHDL to be required because of https://github.com/SpinalHDL/VexRiscv/blob/master/.travis.yml#L48. Hence, `spinalhdl/dev` should fail and `spinalhdl/spinalhdl` should pass. Is this because SpinalHDL is installed as a dependency if not available locally when `sbt compile` is executed?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,64 +1,10 @@
-language: scala
-
-notifications:
-  email:
-    on_success: never
-
-# See 'project/Version.scala'
-scala:
-   - 2.11.6
-
-sbt_args: -no-colors -J-Xss2m
-
-script:
-  - export VEXRISCV_REGRESSION_CONFIG_COUNT=100
-  - export VEXRISCV_REGRESSION_FREERTOS_COUNT=no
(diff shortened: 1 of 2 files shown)


Dolu1990:
Damned, sorry, completely missed that PR :(
Just ping me if you miss feedback !

So basically, most of the time, VexRiscv dev is using some unreleased SpinalHDL version, which is expected to be in the parent directory :
https://github.com/SpinalHDL/VexRiscv/blob/0301ced00095b072332d8d4af081dd18ae3fdcba/build.sbt#L17

So, for vexriscv dev, it might be tricky to use the SpinalHDL docker images.
For master it is ok, but i'm not sure it worth it. as most of the things are done outside master.

PR #64, opened 2019-03-18
MulSimplePlugin

tomverbeure:
Simplest way to implement a multiplier in a VexRiscv: with a 33x33 = 65 bit multiplier...

It's up to the FPGA or ASIC synthesis tool to take care of the rest. And sometimes that gives better results than the current MulPlugin.

Tom


Dolu1990:
Cool, thanks ^^

PR #66, opened 2019-03-23
Add getPipelinedMemoryBusConfig()

tomverbeure:
This is a trivial patch, but the lack of this function while similar ones were present for other busses, made me assume that .toPipelinedMemoryBus() didn't exist...

Tom


Dolu1990:
Thanks ^^

PR #69, opened 2019-04-08
GenMicro with warnings

tomverbeure:
* Creates a configuration with removed WriteBack and Memory stage.
* Add warning when Memory stage is disabled and earlyBranch is false.

### DIFF
--- a/src/main/scala/vexriscv/demo/GenMicroNoCsr.scala
+++ b/src/main/scala/vexriscv/demo/GenMicroNoCsr.scala
@@ -1,0 +1,61 @@
+package vexriscv.demo
+
+import vexriscv.plugin._
+import vexriscv.{plugin, VexRiscv, VexRiscvConfig}
+import spinal.core._
+
+/**
+ * Created by spinalvm on 15.06.17.
+ */
+object GenMicroNoCsr extends App{
+  def cpu() = new VexRiscv(
+    config = VexRiscvConfig(
+      withMemoryStage = false,
+      withWriteBackStage = false,
+      plugins = List(
+        new IBusSimplePlugin(
+          resetVector = 0x80000000l,
+          cmdForkOnSecondStage = false,
+          cmdForkPersistence = false,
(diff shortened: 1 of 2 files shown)


Dolu1990:
Hooo thanks, that's one of the thing i wanted to do, update the small configs with one which use less stages :)
The thing about the configs without all stages is that they aren't so much tested. Not saying they are buggy. but currently there is none of them into the scala tests.
https://github.com/SpinalHDL/VexRiscv/blob/master/src/test/scala/vexriscv/DhrystoneBench.scala
https://github.com/SpinalHDL/VexRiscv/blob/master/src/test/scala/vexriscv/TestIndividualFeatures.scala#L336

Did you tried it into the regressions stuff (src/test/cpp/regression)?

tomverbeure:
I’ve been running CoreMark in them in my coremark tree, but I can also add them to regression.

Tom

Dolu1990:
Hooo cool, so i reccently added a COREMARK=yes option into the linux branch cpp/regression stuff. 
Can you tell me what are the coremark performance it give you ? 
Then i will compaire it with what the cpp/regression is telling me, just to be sure we have the same number, and both ported the coremark benchmark in the same way :)

With benchmark i'm realy carefull XD

tomverbeure:
I have a million different scores for different configurations.

I’m working on a blog post where it analyze the performance impact of CPU architectural features, compiler version, compiler options etc.

Tom

Dolu1990:
Cool :D

Let's me know when it is released :)

Dolu1990:
@tomverbeure I submited two fixes, together they increased Fmax from 100 Mhz to 140 Mhz on  cyclone IV for the GenFullNoMmuNoCache config (for instance) 
Enjoy :D

tomverbeure:
I have done some clock speed testing in the past and the branch predicitor indeed reduced clock speed by a lot.

Good thing that I didn’t start speed testing yet this time. :-)

Tom

PR #74, opened 2019-05-07
Add memory 2 stage

tomverbeure:
This change add an additional memory to the pipeline for those cases where maximum clock is speed is required and data RAM latency is 2.

Right now, you can already do this by setting emitCmdInMemoryStage = false and earlyInjection = false (which are the defaults), but then you have a potential critical path going from execute to the memory.

Most high speed RISC-V CPUs have 2 memory stages.

This has been tested by running coremark, but I haven't run riscv-formal on it.

The change isn't perfect in that the fast version of Mul still injects the final result in WriteBack whereas it could do it in Memory2.

What do you think? Is this worthy of being part of the main tree?

Tom


Dolu1990:
Hi,

So, currently the main issue of having a two stage load/store unit is to handle precise exceptions. Basicaly, if you have a read which throw an exception, the next instruciton should not be executed and should not change any state.

Currently, with the two stage load/Store, the issue is that the exception is managed too late in the pipeline, and maybe a write was already sent to the memory before the faulty read error comeback.

To handle that there would be a solution : 
- Defining the memory regions which have side effect when reading them (peripherals)
- Do not schedule any write or read (on memory regions with side effect), while a read is pending.

That should do it. 
Then about the mul plugin, it could be smarter and stretch istelf from the stage starting from the execute stage.

Else, that pull request is a good feature, if that pass all the test, it can be merged in main, as a DBusSimplePlugin only feature :) 
There is no realy point to update the datacache to spread over more stages.

If you want to run complet regression test, on your own, you can edit : 
https://github.com/SpinalHDL/VexRiscv/blob/master/src/test/scala/vexriscv/TestIndividualFeatures.scala#L384
and 
https://github.com/SpinalHDL/VexRiscv/blob/master/src/test/scala/vexriscv/TestIndividualFeatures.scala#L640
To make the test generating random configuration with the additional stage.

PR #79, opened 2019-07-11
Litex target

mateusz-holenko:
This reworks building emulator a little and adds new target for emulator - LiteX platform.


mateusz-holenko:
@Dolu1990 What do you think about my proposition of adding LiteX target for emulator?

Dolu1990:
Hi,

Thanks for the pull requests, sorry i’m in holiday in a foreign country until friday.
So currently there is to many copy of the machine mode emulator, and that’s great to consolidate it. ( second copy : https://github.com/SpinalHDL/SaxonSoc/tree/Bmb/software/standalone/machineModeSbi/src)

When i will be back i will manage it :)

Dolu1990:
Thanks :)

PR #80, opened 2019-07-24
Fix handling LiteX uart and timer.

mateusz-holenko:
This PR adds fixes to LiteX uart and timer "drivers".

I did my previous implementation in an emulator that failed to reproduce some behaviour. This version is tested in an actual FPGA.


Dolu1990:
Thanks ^^

PR #86, opened 2019-09-20
Short pipeline fixes

xobs:
This patchset fixes a few issues I've experienced with short pipelines refusing to build.

I'm unclear of whether these patches generate a working core without hazards, as I'm still not clear on the internals of vex.  But the resulting file appears to work.


Dolu1990:
Hoo thanks that's usefull, i will review that more in depth.
That's funny that is work for the dcache like this ^^
Basicaly, there is a bit of logic in the datacache which check hazard on the cache memory which could be in part remmoved if it only using a single stage : 
https://github.com/SpinalHDL/VexRiscv/blob/master/src/main/scala/vexriscv/ip/DataCache.scala#L460

More generaly the whole stageA could be removed to reduce stages : 
https://github.com/SpinalHDL/VexRiscv/blob/master/src/main/scala/vexriscv/ip/DataCache.scala#L448

Then i also have to add those configuration the regression tests.

But it seem fine :D

Dolu1990:
Wait, you are using the data cache without memory/writeback stages ?

xobs:
No, I never got it to build the dcache. I'm running with icache only.

Dolu1990:
Thanks ^^

PR #94, opened 2019-10-20
Multiplication Plugin using 16-bit DSPs

rpls:
This adds yet another multiplication plugin variant. The goal here was to only use 16-bit multiplications to implement a 32x32bit multiplier, with the intention to utilize 16-bit DSPs of some FPGAs. I've confirmed that the Icestorm (Yosys+Nextpnr) toolchain will synthesize and place this into the DSP cells of a iCE40 UltraPlus 5K FPGA. Yosys requires the `-dsp` flag for the `synth_ice40` command and will then correctly detect the multiplication operator. 

I originally developed this in the context of [https://github.com/mupq/pqriscv-vexriscv](https://github.com/mupq/pqriscv-vexriscv), but I think others might find this interesting. And I also hope to get some feedback on it.

### DIFF
--- a/src/main/scala/vexriscv/plugin/Mul16Plugin.scala
+++ b/src/main/scala/vexriscv/plugin/Mul16Plugin.scala
@@ -1,0 +1,119 @@
+package vexriscv.plugin
+
+import vexriscv._
+import vexriscv.plugin._
+import spinal.core._
+
+/**
+  * A multiplication plugin using only 16-bit multiplications
+  */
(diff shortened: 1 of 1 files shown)


mithro:
@xobs - I think you will be very interested in this!

Dolu1990:
It pass the regression tests ^^
I will just fix the Div plugin for the configs without execute/memory stage and then merge this pull request
Thanks !

PR #105, opened 2020-01-12
XIP on Murax/iCE40-HX8K breakout improvements

sebastien-riou:
rom bootloader:

- fix to work with latest version of SPI XDR
- minor improvement: check a magic word in flash before jumping to it. This allows to program the hardware, connect J7 and have the execution starting on flash without having to use JTAG.

Murax.scala:

- fix the RAM size to 8KB
- dead code clean up


Dolu1990:
All good, but the magic word is a bit scary to me, as it require the user binary to be specificly patched to run on XIP, isn't it ?

sebastien-riou:
it just requires the user binary to start with "li x31, 0x12340000" ("lui	t6,0x12340" in fact). Sure any image produced so far would not be compatible but that's a minor adaptation and it does make the demo a bit easier to use. 
Without it, we have to use the JTAG to reset the device after connecting J7. With it, the execution starts right after we connect J7. I did not bother to do "debouncing" but that seems to work well enough, it did failed to start so far on my side. Alternatively we could wait for a byte from UART before jumping to flash.

roman3017:
I think the magic start instruction should be mentioned in readme file or one may try otherwise legal binary and spend hours of debugging. But avoiding it would be preferable for me.

sebastien-riou:
Well that's the proof that it is not explained at the right place. I uploaded a new readme putting this first (also correct minor display problems).

roman3017:
It is really nice to see so many improvements.

sebastien-riou:
It would be nice to be able to simulate it as well. I took some steps towards that but I am stuck with a problem in verilator and the iCE40 pads model (I took code from https://github.com/YosysHQ/yosys/blob/master/techlibs/ice40/cells_sim.v ).
How did you simulate this in the past ? Do you still have that setup somewhere ?

PR #109, opened 2020-01-17
missing tcl files for Murax on Arty

sebastien-riou:
also fix the clean target of Murax xipbootloader on HX8K


Dolu1990:
thanks ^^

PR #113, opened 2020-02-11
Murax on arty_a7: fix inferred memory type

japm48:
My Vivado version (2018.3) infers `RAMB18E1`s instead of `RAMB36E1`s, which is probably more area-efficient (or maybe it's the same after optimization). But this breaks a TCL script that assumes the second case; after this patch, it should be compatible with both.

### DIFF
--- a/scripts/Murax/arty_a7/make_vivado_project
+++ b/scripts/Murax/arty_a7/make_vivado_project
@@ -3,7 +3,7 @@
 #cannot rm build because it erase software images that the make file copy there
 #rm -rf ./build
 
-mkdir ./build
+mkdir -p ./build
 
 cd ./build
 vivado -mode batch -source ../make_vivado_project.tcl -notrace
--- a/scripts/Murax/arty_a7/soc_mmi.tcl
+++ b/scripts/Murax/arty_a7/soc_mmi.tcl
@@ -21,7 +21,7 @@ proc swap_bits { bit } {
 # open_run impl_1
 
 # Find all the RAMs, place in a list
-set rams [get_cells -hier -regexp {.*core/system_ram/.*} -filter {REF_NAME =~ RAMB36E1}]
+set rams [get_cells -hier -regexp {.*core/system_ram/.*} -filter {REF_NAME == RAMB36E1 || REF_NAME == RAMB18E1}]
 
 puts "[llength $rams] RAMs in total"
 foreach m $rams {puts $m}
@@ -89,7 +89,7 @@ set i 0
 foreach ram $rams {
     # Get the RAM location
     set loc_val [get_property LOC [get_cells $ram]]
-    regexp -- {(RAMB36_)([0-9XY]+)} $loc_val full ram_name loc_xy
+    regexp {(RAMB.+_)([0-9XY]+)} $loc_val full ram_name loc_xy
 
     set memi [dict create ram $ram loc $loc_xy]
 


Dolu1990:
Hm not sure, the issue about the ram inferation currently is that it is tricky to infer byte mask for write accesses, so SpinalHDL instead just generate one ram for each byte.

So 8 KB/4 => 2 KB per ram => RAMB18E1

Or you are generating with more ram than  8KB ?

japm48:
I am using the default build (`vexriscv.demo.Murax_arty`), so it should be `onChipRamSize = 32 kB`.

So, in the end it's a bit confusing: it generates 4 `RAMB18`s for `system_ram` (`MuraxPipelinedMemoryBusRam`) and, for each of them, 8kbits are actually used, as expected.

But it also generates another 2 `RAMB18`s (don't know what size) for `RegFilePlugin` and I suspect that this shouldn't happen; it also complains about not registering the RAM outputs (for timing), so it's inferring "async" RAMs. It doesn't show it clearly in the synth report, and I had to find in the implemented result where these 2 additional RAMs came from.

You can have a look at the logs: https://gist.github.com/japm48/07084d9fe9706643334aba13b2477370

And this is a relevant table from the synthesis:

```
Block RAM: Preliminary Mapping  Report (see note below)

+----------------------------+-----------------+------------------------+---+---+------------------------+---+---+------------------+--------+--------+
|Module Name                 | RTL Object      | PORT A (Depth x Width) | W | R | PORT B (Depth x Width) | W | R | Ports driving FF | RAMB18 | RAMB36 | 
+----------------------------+-----------------+------------------------+---+---+------------------------+---+---+------------------+--------+--------+
|MuraxPipelinedMemoryBusRam: | ram_symbol0_reg | 1 K x 8(READ_FIRST)    | W |   | 1 K x 8(WRITE_FIRST)   |   | R | Port A and B     | 1      | 0      | 
|MuraxPipelinedMemoryBusRam: | ram_symbol1_reg | 1 K x 8(READ_FIRST)    | W |   | 1 K x 8(WRITE_FIRST)   |   | R | Port A and B     | 1      | 0      | 
|MuraxPipelinedMemoryBusRam: | ram_symbol2_reg | 1 K x 8(READ_FIRST)    | W |   | 1 K x 8(WRITE_FIRST)   |   | R | Port A and B     | 1      | 0      | 
|MuraxPipelinedMemoryBusRam: | ram_symbol3_reg | 1 K x 8(READ_FIRST)    | W |   | 1 K x 8(WRITE_FIRST)   |   | R | Port A and B     | 1      | 0      | 
+----------------------------+-----------------+------------------------+---+---+------------------------+---+---+------------------+--------+--------+
```
Here a screenshot with the implemented design:

Dolu1990:
> But it also generates another 2 RAMB18s (don't know what size) for RegFilePlugin

That's the 1 write 2 read ports of the CPU 32 registers.
Basicaly, the Murax cpu config use syncronus reads for the reg file, which isn't necessarly optimal for FPGA with distributed ram, there is an option to use asycronus reads instead.

> it also complains about not registering the RAM outputs (for timing)

That's normal, no worries about that

> But it also generates another 2 RAMB18s (don't know what size) for RegFilePlugin and I suspect that this shouldn't happen; it also complains about not registering the RAM outputs (for timing), so it's inferring "async" RAMs

Where do you see that it infer asyc ram for the RegFilePlugin ? i don't see that.
but
```
INFO: [Synth 8-6837] The timing for the instance core/system_cpu/RegFilePlugin_regFile_reg_1 (implemented as a Block RAM) might be sub-optimal as no optional output register could be merged into the block ram. Providing additional output register may help in improving timing.
INFO: [Synth 8-6837] The timing for the instance core/system_cpu/RegFilePlugin_regFile_reg_2 (implemented as a Block RAM) might be sub-optimal as no optional output register could be merged into the block ram. Providing additional output register may help in improving timing.
I
```

japm48:
Thanks!

 > Where do you see that it infer asyc ram for the RegFilePlugin ? i don't see that.

My mistake, I confused async and output register.

PR #115, opened 2020-02-24
emulator: Use external hw/common.h from LiteX

mateusz-holenko:
Remove code copied from `hw/common.h` and use the header from the LiteX repository provided
using `LITEX_BASE` environment variable.

Content of `common.h` is now evolving (new functions are added, some are removed) and syncing it between repos would be cumbersome.


Dolu1990:
Sorry i missed it, thanks :)

PR #124, opened 2020-05-19
Add uinstret support.

tomverbeure:


Dolu1990:
Thanks :D

It trigger a Unspecified value parameter uinstretAccess.
I'm totaly fine with just giving a default value to the added xxxAccess in the CsrConfig class to avoid backward compatibility issues :)

Dolu1990:
(default being CsrAccess.NONE)

tomverbeure:
I'll add it!

tomverbeure:
Default value added...

Dolu1990:
Thanks :)

PR #135, opened 2020-08-30
Add support for big endian byte ordering

zeldin:
I have tested this on an iCE40 HX8K breakout board.

Big-endian support for the toolchain can be found [here](https://github.com/zeldin/binutils-gdb/tree/riscv-bigendian) and [here](https://github.com/zeldin/gcc/tree/riscv-bigendian).

Interrestingly enough, `nextpnr-ice40` reports a higher fMax (73.09 MHz w/o RVC, 66.64 MHz with RVC) for big endian Murax than for little endian (72.35 MHz w/o RVC, 62.09 MHz with RVC), although I don't know why big endian would be faster...

### DIFF
--- a/src/main/scala/vexriscv/demo/Murax.scala
+++ b/src/main/scala/vexriscv/demo/Murax.scala
@@ -52,8 +52,8 @@ case class MuraxConfig(coreFrequency : HertzNumber,
 
 
 object MuraxConfig{
-  def default : MuraxConfig = default(false)
-  def default(withXip : Boolean) =  MuraxConfig(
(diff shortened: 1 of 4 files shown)


Dolu1990:
Hi,
Thanks :D
So, don't worry yet about the travis failure, that's a flow issue, i will try to fix that (maybe the fix is just to wait something to be fixed in the cloud XD)

zeldin:
@Dolu1990 Yeah, I looked at the Travis log and figured that it couldn't possibly be my fault...  :smile:

Dolu1990:
thanks ^^

PR #136, opened 2020-09-06
Add support for RV32E in RegFilePlugin

zeldin:
The RV32E extension removes registers x16-x31 from the ISA.  This
is useful when compiling with -mem2reg to save on BRAMs.  On iCE40
HX8K this option saves 1285 LC:s, which also improves the routing
situation, when using -mem2reg.

Note that the illegal instruction exception required by the RV32E
specification for accesses to registers x16-x31 is not implemented.

----

The missing exception could probably be implemented, but I'm afraid I'm not savvy enough in SpinalHDL to pull it off...  :sweat_smile:

### DIFF
--- a/src/main/scala/vexriscv/plugin/RegFilePlugin.scala
+++ b/src/main/scala/vexriscv/plugin/RegFilePlugin.scala
@@ -18,6 +18,7 @@ class RegFilePlugin(regFileReadyKind : RegFileReadKind,
                     writeRfInMemoryStage : Boolean = false,
                     readInExecute : Boolean = false,
                     syncUpdateOnStall : Boolean = true,
+                    rv32e : Boolean = false,
                     withShadow : Boolean = false //shadow registers aren't transition hazard free
                    ) extends Plugin[VexRiscv] with RegFileService{
   import Riscv._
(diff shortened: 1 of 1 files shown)


Dolu1990:
thanks ^^

Hmm to me, a RV32 instruction on a RV32E CPU should trigger a "Illegal instruction" (code 2). So basicaly, it would be about changing the instruction patterns that are reconized as instruction in each plugin which use the register file.

ex :
ALU => https://github.com/SpinalHDL/VexRiscv/blob/master/src/main/scala/vexriscv/plugin/IntAluPlugin.scala#L54
ADD kexword => https://github.com/SpinalHDL/VexRiscv/blob/24b676ce30eb47c20b6fb0ac1c0b6f980abba7b4/src/main/scala/vexriscv/Riscv.scala#L31
Which should depend on RV32[E] to fit the proper encoding space.

zeldin:
@Dolu1990 Hm, so what you are saying is change the most significant bit of rd/rs1/rs2 in each of those patterns from a "-" to a "0"?  I guess that should work, but it seems a bit messy to have two variants of each pattern when the difference is always the same...

PR #140, opened 2020-10-04
Add ASID field to SATP

bunnie:
ASID field is missing from the SATP which causes compatibility
issues with Xous.

While this patch resolves the Xous issue, it has not been tested
on Linux. If the ASID omission was intentional because it caused
problems for Linux, then this PR should be rejected.


Dolu1990:
ASID was intentionaly not implemented as not required by the spec, but as this is only 9 bits, let's keep it in ^^
The travis regression run some linux test, so seem good. Thanks :)

Can you just move from master to dev branch ?

bunnie:
Closing this request as it is moved to `dev` branch as requested.

PR #141, opened 2020-10-04
Dev asid

bunnie:
Note - this commit is moved to 'dev' per request of @Dolu1990 

I think I've done the move correctly; let me know if I messed something up in terms of merging into the dev branch.

Remainder of message as original request --

ASID field is missing from the SATP which causes compatibility
issues with Xous.

While this patch resolves the Xous issue, it has not been tested
on Linux. If the ASID omission was intentional because it caused
problems for Linux, then this PR should be rejected.


Dolu1990:
Thanks, all good ^^

PR #145, opened 2020-10-20
Update big endian instruction encoding

zeldin:
Between draft-20181101-ebe1ca4 and draft-20190622-6993896 of the
RISC-V Instruction Set Manual, the wording was changed from requiring
"natural endianness" of instruction parcels to require them to be
little endian.

Update the big endian instruction pipe to reflect the newer requirement.

Updated toolchains can be found here:
https://github.com/zeldin/riscv-binutils-gdb/commits/big-endian
https://github.com/zeldin/riscv-gcc/commits/big-endian


Dolu1990:
Thanks :)

PR #146, opened 2020-10-22
Regression Fix: Only Check IO Region

elliott-wen:


Dolu1990:
Basicaly as currently the data cache is write through, all the bus write can be checked against the reference model, which allow to catch early bugs.

It would be better (as a patch) to add a config in the makefile to specify if the dbus is write back or write through, to  enable / disable the write check on cached region. (default write-through) :)

Dolu1990:
any way nice catch XD

PR #147, opened 2020-10-27
Physical Memory Protection (PMP) plugin

lindemer:
This plugin enables memory protection features using either the NA4 or NAPOT addressing modes. I've added a configuration file for Zephyr targets that will support its new RISC-V userspace features (M and U mode support, 16 PMP regions).

I'm posting this as a draft to get some early feedback. I still need to write a couple regression tests. One specific question: _The PMP registers have a lock bit which prevents all writes (even in M-mode) until reset. Can this be implemented with the current CSR interface?_


Dolu1990:
It look nice and clean :)

> The PMP registers have a lock bit which prevents all writes (even in M-mode) until reset. Can this be implemented with the current CSR interface?

Hmm maybe like this : 


```scala
val lockSet = False
val lock = RegInit(False) setWhen(lockSet)
csrService.w(CSR, BIT, lockSet)
```

Basicaly, this allow to set, but not to clear, is it what you want ? (seem to fit the privilege spec)

lindemer:
Okay thanks :) I think there still might be an issue with this approach. The CSR fields are declared as `Reg` so the user can read and write them. The internal state (`regionStart`, `regionEnd`, `locked`, `valid`) also has to be `Reg` so that it persists even if the user attempts to write the register again after setting the `L` bit.

Since both the CSR and the internal state are declared `Reg`, does that mean it will take 2 CPU cycles for the PMP restrictions to take effect? (Maybe this is acceptable anyways.) Also, since the `L` bit is now write-only, will an exception occur if the user attempts to read the entire register?

lindemer:
Okay I think this can actually be solved just by separating the read and write permissions on the CSR. Now, there's a layer of logic between what the user can write and the actual registers storing the values. Once the lock bit has been set, attempts to write shouldn't change the internal state of the CSR.

However, this isn't compiling now. Any idea on what's happening with this error?
```
[error] /home/nebula/VexRiscv/src/main/scala/vexriscv/plugin/PmpPlugin.scala:197:37: overloaded method value apply with alternatives:
[error]   [T <: spinal.core.Data](oneHot: IndexedSeq[spinal.core.Bool], inputs: spinal.core.Vec[T])T <and>
[error]   [T <: spinal.core.Data](oneHot: spinal.core.BitVector, inputs: spinal.core.Vec[T])T <and>
[error]   [T <: spinal.core.Data](oneHot: IndexedSeq[spinal.core.Bool], inputs: Iterable[T])T <and>
[error]   [T <: spinal.core.Data](oneHot: spinal.core.BitVector, inputs: Seq[T])T
[error]  cannot be applied to (spinal.core.Vec[spinal.core.Bool], scala.collection.mutable.ArrayBuffer[vexriscv.plugin.PmpRegister])
[error]           port.bus.rsp.allowRead := MuxOH(OHMasking.first(hits), pmps).r
```
It looks like it's complaining that the input types don't match, but as far as I can tell they actually do.

Dolu1990:
Hmm maybe try to cast pmps.toSeq , same for the first(...)
Sometime the scala colections can be tricky, i'm not sure i mastered them properly XD

lindemer:
Looks like it's not such an easy fix... `MuxOH` was happy in the first commit because `PmpRegister` was declared as `extends Bundle`. After I added the locking logic it doesn't compile with `extends Bundle` anymore, and that seems to break `MuxOH`. Is there something else I can extend the class with to make it work with `MuxOH`?

Dolu1990:
hmm i would need to try it, can you push the broken thing for me to try ?

lindemer:
Yup! I pushed the latest commit 15 minutes ago, thanks for looking into it :)

Dolu1990:
This should be fine them : 

```scala
          port.bus.rsp.allowRead := MuxOH(OHMasking.first(hits), pmps.map(_.region.r))
          port.bus.rsp.allowWrite := MuxOH(OHMasking.first(hits), pmps.map(_.region.w))
          port.bus.rsp.allowExecute := MuxOH(OHMasking.first(hits), pmps.map(_.region.x))
```

lindemer:
Thanks, that solves the `MuxOH` problem but now every value in `PmpRegister` is throwing a hierarchy violation. In the documentation it says this is usually due to using `Bundle` incorrectly but I don't have any declared here.

Dolu1990:
So
basicaly, instead of having PmpRegister as a Component, instead have them as an Area :)
That should fix it.

lindemer:
Ah sorry for all the debugging questions, but when I change it from `PmpRegister() extends Component` to `PmpRegister() extends Area` I get this error, but it doesn't really point to a specific line of code:
```
[info] [Progress] at 1.331 : Checks and transforms
[error] Exception in thread "main" java.lang.AssertionError: assertion failed
[error] 	at scala.Predef$.assert(Predef.scala:156)
[error] 	at spinal.core.package$.assert(core.scala:405)
[error] 	at spinal.core.internals.SwitchStatement.bitVectorNormalize$1(Statement.scala:464)
[error] 	at spinal.core.internals.SwitchStatement.normalizeInputs(Statement.scala:484)
[error] 	at spinal.core.internals.PhaseNormalizeNodeInputs$$anonfun$impl$30.apply(Phase.scala:1278)
[error] 	at spinal.core.internals.PhaseNormalizeNodeInputs$$anonfun$impl$30.apply(Phase.scala:1274)
[error] 	at spinal.core.internals.TreeStatement$$anonfun$walkStatements$2.apply(Statement.scala:264)
[error] 	at spinal.core.internals.TreeStatement$$anonfun$walkStatements$2.apply(Statement.scala:262)
[error] 	at spinal.core.internals.ScopeStatement.foreachStatements(Statement.scala:136)
[error] 	at spinal.core.internals.WhenStatement.foreachStatements(Statement.scala:388)
[error] 	at spinal.core.internals.TreeStatement$class.walkStatements(Statement.scala:262)
[error] 	at spinal.core.internals.WhenStatement.walkStatements(Statement.scala:382)
[error] 	at spinal.core.internals.ScopeStatement$$anonfun$walkStatements$1.apply(Statement.scala:143)
[error] 	at spinal.core.internals.ScopeStatement$$anonfun$walkStatements$1.apply(Statement.scala:141)
[error] 	at spinal.core.internals.ScopeStatement.foreachStatements(Statement.scala:136)
[error] 	at spinal.core.internals.ScopeStatement.walkStatements(Statement.scala:141)
[error] 	at spinal.core.internals.PhaseContext$$anonfun$walkStatements$1.apply(Phase.scala:148)
[error] 	at spinal.core.internals.PhaseContext$$anonfun$walkStatements$1.apply(Phase.scala:148)
[error] 	at spinal.core.internals.GraphUtils$.walkAllComponents(Misc.scala:261)
[error] 	at spinal.core.internals.PhaseContext.walkStatements(Phase.scala:148)
[error] 	at spinal.core.internals.PhaseNormalizeNodeInputs.impl(Phase.scala:1274)
[error] 	at spinal.core.internals.PhaseContext.doPhase(Phase.scala:196)
[error] 	at spinal.core.internals.SpinalVerilogBoot$$anonfun$singleShot$10.apply(Phase.scala:2373)
[error] 	at spinal.core.internals.SpinalVerilogBoot$$anonfun$singleShot$10.apply(Phase.scala:2371)
[error] 	at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
[error] 	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
[error] 	at spinal.core.internals.SpinalVerilogBoot$.singleShot(Phase.scala:2371)
[error] 	at spinal.core.internals.SpinalVerilogBoot$.apply(Phase.scala:2306)
[error] 	at spinal.core.Spinal$.apply(Spinal.scala:322)
[error] 	at spinal.core.SpinalConfig.generate(Spinal.scala:146)
[error] 	at spinal.core.SpinalVerilog$.apply(Spinal.scala:340)
[error] 	at vexriscv.demo.GenZephyr$.delayedEndpoint$vexriscv$demo$GenZephyr$1(GenZephyr.scala:84)
[error] 	at vexriscv.demo.GenZephyr$delayedInit$body.apply(GenZephyr.scala:8)
[error] 	at scala.Function0$class.apply$mcV$sp(Function0.scala:34)
[error] 	at scala.runtime.AbstractFunction0.apply$mcV$sp(AbstractFunction0.scala:12)
[error] 	at scala.App$$anonfun$main$1.apply(App.scala:76)
[error] 	at scala.App$$anonfun$main$1.apply(App.scala:76)
[error] 	at scala.collection.immutable.List.foreach(List.scala:392)
[error] 	at scala.collection.generic.TraversableForwarder$class.foreach(TraversableForwarder.scala:35)
[error] 	at scala.App$class.main(App.scala:76)
[error] 	at vexriscv.demo.GenZephyr$.main(GenZephyr.scala:8)
[error] 	at vexriscv.demo.GenZephyr.main(GenZephyr.scala)
[error] Nonzero exit code returned from runner: 1
[error] (Compile / runMain) Nonzero exit code returned from runner: 1
[error] Total time: 4 s, completed Oct 29, 2020 11:51:36 AM
```

Dolu1990:
I think it is because     val a = Reg(UInt) init(0)   has no bit width set
Can you try      val a = Reg(UInt(?_? bits) init(0)

lindemer:
Okay everything compiles now, thanks :) Now I'm wondering:

1. Should the pipeline be flushed or stalled after writing to the PMP CSRs? If the CSRs are updated in the _writeback_ stage, then 5 instructions will be fetched and 2 will pass the _memory_ stage before the PMP restrictions are updated. (The PMP restrictions will take effect 1 cycle after the CSR is written due to the way I've implemented it.)
2. Could you point me to a good starting point to write some regression tests for this plugin?

Dolu1990:
> Should the pipeline be flushed or stalled after writing to the PMP CSRs

So basicaly, CSR a special case
When a CSR access is in the execut stage, the CsrPlugin wait that the memory stage and write back stage are empty before executing the CSR.
CSR are executed in the execute stage.

Actualy, when the SATP register is written, the CsrPlugin jump at PC+4 to flush the pipe. the it is done isn't perfect, as it do not allow your plugin to ask the reschedule. That should be something to add in the CsrService (saying if a given CSR need to flush the pipeline) and in your case you would need that flush on each PMP CSR writes.
(https://github.com/SpinalHDL/VexRiscv/blob/master/src/main/scala/vexriscv/plugin/CsrPlugin.scala#L657)

> Could you point me to a good starting point to write some regression tests for this plugin?

That's a bit a mess, VexRiscv realy growed over time, from an inexperianced ground.
So, you can write regression asm in : 
https://github.com/SpinalHDL/VexRiscv/tree/master/src/test/cpp/raw

Then add them to : 
https://github.com/SpinalHDL/VexRiscv/blob/master/src/test/cpp/regression/main.cpp#L3872

By removing the .withRiscvRef() you can disable the Riscv golden model, which can be usefull for dev. But ultimately, the behaviour of the PMP could be added to the golden model.

lindemer:
@Dolu1990 Thanks for the advice, I'll get back to you soon :)

lindemer:
> So basicaly, CSR a special case. When a CSR access is in the execut stage, the CsrPlugin wait that the memory stage and write back stage are empty before executing the CSR. CSR are executed in the execute stage.

Is this true for all CSRs or just SATP? I can't seem to find where this halt occurs in `CsrPlugin.scala`. 

> That should be something to add in the CsrService (saying if a given CSR need to flush the pipeline) and in your case you would need that flush on each PMP CSR writes.

For now I've copied the logic into `PmpPlugin.scala`. To be honest I struggled to figure out how to do what you suggested since I'm not too familiar with Scala.

shufps:
impressive - it looks so much cleaner than my own pmp plugin ... but I'm not sure if it conforms to the official PMP proposal

Dolu1990:
>Is this true for all CSRs or just SATP? I can't seem to find where this halt occurs in CsrPlugin.scala.

For all CSR :
https://github.com/SpinalHDL/VexRiscv/blob/dev/src/main/scala/vexriscv/plugin/CsrPlugin.scala#L1074

> Edit: Aha I guess you mean this. 

RIght ^^ 

> For now I've copied the logic into PmpPlugin.scala. To be honest I struggled to figure out how to do what you suggested since I'm not too familiar with Scala.

It isn't a obvious thing to do, no worries.
Check this : 
https://github.com/SpinalHDL/VexRiscv/blob/dev/src/main/scala/vexriscv/plugin/CsrPlugin.scala#L710

This is basicaly how the CSR plugin enforce that SATP write flush the pipeline and redo the next instruction, basicaly the same could be done for PMP. 
So basicaly, we would need a way to specify when the current CSR require a resquedule of the next instruction, via  an additional primitive into CsrInterface

lindemer:
@shufps Thanks 👽  
@Dolu1990 What's the easiest way to add my regression test to the main test file and run it? I don't think this plugin is compatible with an MMU the way it's currently implemented, so I guess I can't just add PMP to the golden model.

lindemer:
@Dolu1990 Could you give me a quick explainer on what's going on on `TestIndividualFeatures.scala`? I want to add my PMP regression test but I need to make sure that it only runs if (1) U-mode is implemented and (2) no MMU is present. It's a bit unclear what the `VexRiscvUniverse` and `VexRiscvDimension` classes are doing because this file doesn't have any comments.

lindemer:
I managed to add the PMP test to the main regression suite, but it's currently failing. To reproduce:
```
export VEXRISCV_REGRESSION_CONFIG_LINUX_RATE=0
export VEXRISCV_REGRESSION_CONFIG_MACHINE_OS_RATE=1
sbt "testOnly vexriscv.TestIndividualFeatures"
```
The output looks like this:
```
 pc missmatch 8000006c should be 80000010
FAIL pmp at PC=8000006c REF PC=80000010 REF I=341f1073 time=398
 pc missmatch 8000006c should be 80000010
FAIL pmp at PC=8000006c REF PC=80000010 REF I=341f1073 time=396
...
```
The error occurs on the first attempt to write one of the new CSRs (`csrw pmpcfg0, t5`). I tried removing the following from the plugin but it had no effect:
```
execute.arbitration.flushNext := True
redoInterface.valid := True
```
I noticed there are a bunch of `nop`s in the other handwritten regression tests. I tried adding some of those and it does change the error but it doesn't get rid of it. Any ideas on how to fix this?

Dolu1990:
@lindemer 

>  I don't think this plugin is compatible with an MMU the way it's currently implemented

Right, basicaly, that's one weakness of the model, is that actualy all the memory address check and translation should be inside a single plugin. To have a system with Pmp and MMU you would need to have them both merged.

> so I guess I can't just add PMP to the golden model.

Why not ?
If mmu is disabled in csr it will not trigger it.

> The output looks like this:

Did you added the PMP software model in the golden model ?

lindemer:
> Did you added the PMP software model in the golden model ?

No and I'm not sure how to do this. Is there a similar plugin already in it that I can use as a guide?

Dolu1990:
Basicaly, in the regression testbench, this is the software model of the MMU : 
https://github.com/SpinalHDL/VexRiscv/blob/master/src/test/cpp/regression/main.cpp#L428

v2p for virtual to physical.

It return true when there is an exception.

So when the pmp is enabled, it should first check the PMP region, and eventialy return true if it should create an exception.

You will also have to implement the read/write of the PMP CSR, see https://github.com/SpinalHDL/VexRiscv/blob/master/src/test/cpp/regression/main.cpp#L529 and futher ^^

lindemer:
I added PMP to the golden model, but this error is persisting when I attempt to do a CSR write to one of the PMP registers:
```
 pc missmatch 80000080 should be 80000070
FAIL pmp at PC=80000080 REF PC=80000070 REF I=191c02b7 time=378
 pc missmatch 80000080 should be 80000070
FAIL pmp at PC=80000080 REF PC=80000070 REF I=191c02b7 time=392
```
Now, if I comment out the following lines in the PMP plugin it does run the test:
```
execute.arbitration.flushNext := True
redoInterface.valid := True
```
As far as I can tell I'm flushing the pipeline and jumping back two instructions exactly as it's done for the SATP register ([https://github.com/SpinalHDL/VexRiscv/blob/master/src/main/scala/vexriscv/plugin/CsrPlugin.scala#L657](https://github.com/SpinalHDL/VexRiscv/blob/master/src/main/scala/vexriscv/plugin/CsrPlugin.scala#L657)). 

Any idea why the PC is still getting out of sync? There doesn't appear to be any special PC handler for writing the SATP CSR in the main regression test file.

Dolu1990:
hmm, can you merge VexRiscv master back into your branch and try again ? i fixed one bug in vex which could maybe interfer : 
https://github.com/SpinalHDL/VexRiscv/commit/e0ae46e79491383e1c7caa3e438b4b504d229f86

(For sanity XD)

Then, have you tried to use the wave ? Could i take a look ?

lindemer:
Yes please do! Sorry not sure what the wave is 😅 Merging master didn't fix it unfortunately.

Dolu1990:
So basicaly, when you run the regression, add the TRACE=yes arguements, and it will generate a .vcd file for each test
this show all the signal of the simulation in a file.

lindemer:
Got it, but the `.vcd` files are several MiB. Do you want me to commit a specific one for you so you can look at it? Otherwise you can generate them with:
```
sbt "runMain vexriscv.demo.GenZephyr"
cd src/test/cpp/regression
make clean run IBUS=CACHED DBUS=CACHED CSR=yes MMU=no PMP=yes DEBUG_PLUGIN=no MUL=no DIV=no REDO=2 TRACE=yes
```

Dolu1990:
Do not commit any file on the git which is "big" (not text) ^^
As it will make the history heavy

lindemer:
Okay, would you mind generating the waveform on your end and taking a look at it?

Dolu1990:
I just tried, but the branch seem quite broken, most test were failing, and for the pmp i get something else : 
FAIL pmp at PC=       0 REF PC=80000000 REF I=97 time=20000
Other tests where still passing for you ?

lindemer:
When I run it like this all tests pass except for PMP:
```
export VEXRISCV_REGRESSION_CONFIG_LINUX_RATE=0
export VEXRISCV_REGRESSION_CONFIG_MACHINE_OS_RATE=1
sbt "testOnly vexriscv.TestIndividualFeatures"
```
But for some reason that other command causes Dhrystone tests to fail. Maybe my `GenZephyr.scala` configuration is causing the problem... I'll get back to you.

lindemer:
@Dolu1990 It doesn't look like my PR introduced this issue. I tried this on the latest upstream code and got the same errors:
```
sbt "runMain vexriscv.demo.GenFull"
cd src/test/cpp/regression
make clean run IBUS=CACHED DBUS=CACHED CSR=yes MMU=no DEBUG_PLUGIN=no MUL=no DIV=no REDO=2 TRACE=yes
```

lindemer:
So I've added support for TOR addressing and I've verified that the plugin itself works according to the specification by running my assembly test manually on the `GenZephyr.scala` configuration. This result can be replicated by following the GDB/OpenOCD guide in the main README.md file. Register `x28` will contain `0xc0ffee` if all tests pass.

**Notes**
- The tests still fail when executed from `TestIndividualFeatures.scala`. I've temporarily disabled the reference model for the PMP tests, but I'm still getting error codes for some reason. 
- I've removed the pipeline flush after the PMP CSR writes. I think this is overkill for realistic use cases. In Zephyr for example, all 20 PMP registers are stored/restored on every context switch. Flushing the pipeline will therefore add 40 CPU cycles to every context switch. The tradeoff is that it is hypothetically possible to fetch two instructions from protected memory immediately after changing the PMP settings. I don't think this is a vulnerability - or at least I don't see how it could be practically exploited.

lindemer:
@Dolu1990 This plugin passes the regression test suite now. You can force the PMP tests by setting `MACHINE_OS_RATE` to 1 and `LINUX_RATE` to 0. A few caveats:
- The tests fail with `IBusSimplePlugin` and `DBusSimplePlugin` so I've added a new flag in the test script `CACHE_ALL` to force the cached versions when running the PMP tests. Not sure exactly why this is, but maybe those plugins should be fixed in a separate PR.
- I don't think it makes sense to add PMP to the golden model, so I just disabled the reference for those tests. The actual SpinalHDL implementation will be much simpler than the C++ reference, so I think this would just be an exercise in debugging the reference model. I could just add more assembly tests to improve coverage if you think that's necessary.
- There's no pipeline flush after the PMP CSR writes anymore. This seems unnecessary because the only exceptions it would catch are situations where software for some reason made the very next instruction non-executable.
What do you think?

Dolu1990:
Hi ^^

> You can force the PMP tests by setting MACHINE_OS_RATE to 1 and LINUX_RATE to 0. A few caveats:

Hoow, why not having a new flag  instead ? like PMP_ENTRIES=0 (to disable it) and PMP_ENTRIES=Number to enable it ?
Basicaly, the number of pmp entry in RISC-V is free, so the golden model will need to know it to be properly configured.

> I don't think it makes sense to add PMP to the golden model,

Ahhh, basicaly, the golden model is just there to catch bugs as soon as they happen, without you software having to be involved into the detection of the issue. It isn't required.

> There's no pipeline flush after the PMP CSR writes anymore. This seems unnecessary because the only exceptions it would catch are situations where software for some reason made the very next instruction non-executable.

Right, in practice that probably a never hit case.

> Since PMP is now enabled for the MachineOS

This shoudn't be the default i would say. realy need to add a extra configuration flag to the makefile in case there is a PMP

lindemer:
> This shoudn't be the default i would say. realy need to add a extra configuration flag to the makefile in case there is a PMP

Okay I removed PMP from `MachineOs` and added a new position called `Secure` with its own environment variable called `VEXRISCV_REGRESSION_CONFIG_SECURE_RATE`. I guess this should have its own test in the CI suite? (And the existing tests would have to set that variable to 0.)

>  like PMP_ENTRIES=0 (to disable it) and PMP_ENTRIES=Number to enable it ?

This would be tricky since the assembly tests assume 16 PMP registers. I doubt we'd find new bugs by varying the number of registers in the tests.

**Port to LiteX / Zephyr**
I'd really like to get this PMP plugin added to the configurations found in the [LiteX source code](https://github.com/litex-hub/pythondata-cpu-vexriscv) so Zephyr [and Tock](https://github.com/tock/tock/pull/2203) can use it.

- Do you need me to make any more changes to merge this PR?
- Do you have any advice on adding this to LiteX? It looks like the repository is somehow automated [here](https://github.com/litex-hub/pythondata-auto/blob/4d7119eceb41bcb2b5e745d024e7c8a32215c196/modules.ini#L69).

Dolu1990:
Hi,

>  I guess this should have its own test in the CI suite? 

Yes

> This would be tricky since the assembly tests assume 16 PMP registers. I doubt we'd find new bugs by varying the number of registers in the tests

Ahhh ok, so PMP=yes is good enough then ^^

> Do you need me to make any more changes to merge this PR?

I think we can merge it, if this is ok for you :)

> Do you have any advice on adding this to LiteX? It looks like the repository is somehow automated here.

As far i know, this is project used to generate the different config and cache them : 
https://github.com/litex-hub/pythondata-cpu-vexriscv/tree/master/pythondata_cpu_vexriscv/verilog

So updating it would update the VexRiscv used in the baremetal SoCs 

The error you have is related to : 

https://github.com/litex-hub/pythondata-cpu-vexriscv/blob/master/pythondata_cpu_vexriscv/verilog/src/main/scala/vexriscv/GenCoreDefault.scala#L234

What did you modified exactly in the setup ?

lindemer:
Great :) Let's merge it! I posted an [Issue](https://github.com/SpinalHDL/VexRiscv/issues/156) with a more detailed explanation. It looks like something in SpinalHDL 1.4.0 broke the LiteX build script.

PR #148, opened 2020-11-01
Draft plugin to support the Zbp (permutation) subset of B.

rdolbeau:
This plugin adds support for the Zbp (permutation) subset of the [B (bit manipulation)](https://github.com/riscv/riscv-bitmanip) extension.
It was written against commit 5f0c7a7, which is the one currently used by https://github.com/litex-hub/pythondata-cpu-vexriscv  to generate cores for https://github.com/litex-hub/linux-on-litex-vexriscv.

This pull request is probably not ready for merging at it didn't receive much testing. It's more a notification that the code exists to anyone would could be interested. The code was written by-the-book and not tuned in any way. It has only been tested as part of linux-on-litex-vexriscv on an Artix-7 XC7A35T-1.

### DIFF
--- a/src/main/scala/vexriscv/plugin/BitManipZbpPlugin.scala
+++ b/src/main/scala/vexriscv/plugin/BitManipZbpPlugin.scala
@@ -1,29 +1,42 @@
-/*
- * Copyright (c) 2020 Romain Dolbeau <romain.dolbeau@european-processor-initiative.eu>
- * MIT License
- * See the LICENSE file at the top level of this software distribution for details.
- */
-
+// WARNING: this is auto-generated code!
+// See https://github.com/rdolbeau/VexRiscvBPluginGenerator/
 package vexriscv.plugin
 import spinal.core._
 import vexriscv.{Stageable, DecoderService, VexRiscv}
-
-/**
-  * This plugin implements the Zbp extension (subset of the 'B' bit manipulation extension)
-  * using V0.92 <https://github.com/riscv/riscv-bitmanip/blob/master/bitmanip-0.92.pdf>
-  * for RV32
-  */
 object BitManipZbpPlugin {
-  // enumeration for the bitwise operations
-  object BitManipZbpBitwiseCtrlEnum extends SpinalEnum(binarySequential){
-    val XNOR, ORN, ANDN = newElement()
-  }
-  // enumeration for the added operation (BITWISE is sub-selected by BitManipZbpBitwiseCtrlEnum)
-  object BitManipZbpCtrlEnum extends SpinalEnum(binarySequential){
-    val ROR, ROL, GREV, GORC, PACK, PACKU, PACKH, SHFL, UNSHFL, BITWISE = newElement()
-  }
-   object BITMANIPZBP_CTRL extends Stageable(BitManipZbpCtrlEnum())
-   object BITMANIPZBP_BITWISE_CTRL extends Stageable(BitManipZbpBitwiseCtrlEnum())
+	object BitManipZbpCtrlbitwiseEnum extends SpinalEnum(binarySequential) {
+		 val CTRL_ANDN, CTRL_ORN, CTRL_XNOR = newElement()
+	}
+	object BitManipZbpCtrlrotationEnum extends SpinalEnum(binarySequential) {
+		 val CTRL_ROL, CTRL_ROR = newElement()
+	}
+	object BitManipZbpCtrlgrevrocEnum extends SpinalEnum(binarySequential) {
+		 val CTRL_GORC, CTRL_GREV = newElement()
+	}
+	object BitManipZbpCtrlshuffleEnum extends SpinalEnum(binarySequential) {
+		 val CTRL_SHFL, CTRL_UNSHFL = newElement()
+	}
+	object BitManipZbpCtrlpackEnum extends SpinalEnum(binarySequential) {
+		 val CTRL_PACK, CTRL_PACKH, CTRL_PACKU = newElement()
+	}
+	object BitManipZbpCtrlxpermEnum extends SpinalEnum(binarySequential) {
+		 val CTRL_XPERMdotB, CTRL_XPERMdotH, CTRL_XPERMdotN = newElement()
+	}
+	object BitManipZbpCtrlgrevorcEnum extends SpinalEnum(binarySequential) {
+		 val CTRL_GORC, CTRL_GREV = newElement()
+	}
+	object BitManipZbpCtrlEnum extends SpinalEnum(binarySequential) {
+		 val CTRL_bitwise, CTRL_rotation, CTRL_grevroc, CTRL_shuffle, CTRL_pack, CTRL_xperm, CTRL_grevorc = newElement()
+	}
+	object BitManipZbpCtrlbitwise extends Stageable(BitManipZbpCtrlbitwiseEnum())
+	object BitManipZbpCtrlrotation extends Stageable(BitManipZbpCtrlrotationEnum())
+	object BitManipZbpCtrlgrevroc extends Stageable(BitManipZbpCtrlgrevrocEnum())
+	object BitManipZbpCtrlshuffle extends Stageable(BitManipZbpCtrlshuffleEnum())
+	object BitManipZbpCtrlpack extends Stageable(BitManipZbpCtrlpackEnum())
+	object BitManipZbpCtrlxperm extends Stageable(BitManipZbpCtrlxpermEnum())
+	object BitManipZbpCtrlgrevorc extends Stageable(BitManipZbpCtrlgrevorcEnum())
+	object BitManipZbpCtrl extends Stageable(BitManipZbpCtrlEnum())
+// Prologue
 
    // function implementing the semantic of 32-bits generalized reverse
    def fun_grev( a:Bits, b:Bits ) : Bits = {
@@ -46,9 +59,9 @@ object BitManipZbpPlugin {
 
    // helper function for the implementation of the generalized shuffles
    def fun_shuffle32_stage(src:Bits, maskL:Bits, maskR:Bits, N:Int) : Bits = {
-    val x = src & ~(maskL | maskR)
-    val x2 = x | ((src |<< N) & maskL) | ((src |>> N) & maskR);
-    return x2;
+       val x = src & ~(maskL | maskR)
+       val x2 = x | ((src |<< N) & maskL) | ((src |>> N) & maskR);
+       x2 // return value
    }
    // function implementing the semantic of 32-bits generalized shuffle
    def fun_shfl32(a:Bits, b:Bits) : Bits = {
@@ -57,7 +70,7 @@ object BitManipZbpPlugin {
        val x2 = ((b&B"32'x00000004")===B"32'x00000004") ? fun_shuffle32_stage(x1, B"32'x0F000F00", B"32'x00F000F0", 4) | x1;
        val x3 = ((b&B"32'x00000002")===B"32'x00000002") ? fun_shuffle32_stage(x2, B"32'x30303030", B"32'x0C0C0C0C", 2) | x2;
        val x4 = ((b&B"32'x00000001")===B"32'x00000001") ? fun_shuffle32_stage(x3, B"32'x44444444", B"32'x22222222", 1) | x3;
-       return x4;
+       x4 // return value
    }
    // function implementing the semantic of 32-bits generalized unshuffle
    def fun_unshfl32(a:Bits, b:Bits) : Bits = {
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi,

Cool, thanks, i'm not aware of another B implementation on Vex :)
About testing, if you want, i can show you where to implement / add test in VexRiscv repo  (without soc) and with automated regressions ?

rdolbeau:
I must confess I don't really know how to use the pure VexRiscv repo - I relied exclusively on linux-on-litex-vexrsicv to get a 'usable' SoC as most of the work was there already (I only did the board support, really). So it wasn't even written/tested against HEAD, but an older commit. VexRiscv does run the Chacha20 crypto algorithm about 30% faster with 'rori' than without, so for some use cases it might be worth the extra area.

I'm a SW guy not a HW guy so I don't understand all of the HW side... It's also a proof of sort that SpinalHDL is somewhat easy to use for newcomers!

If you have a pointer to some documentation on testing I'm interested, but no promises - this is an offshoot of a hobby project (for which I got the board) so not a lot of time available for it ;-)

Dolu1990:
> I must confess I don't really know how to use the pure VexRiscv repo

No worries ^^


> I'm a SW

I was suspecting it seeing : 
https://github.com/SpinalHDL/VexRiscv/pull/148/files#diff-5986ab15b81931bcedf2af0c91287519799eff825f6ee537b8405b63b83faaaeR54

Which is a very very software way to do it :)

> It's also a proof of sort that SpinalHDL is somewhat easy to use for newcomers!

Heretic !!!

> If you have a pointer to some documentation on testing I'm interested, but no promises - this is an offshoot of a hobby project (for which I got the board) so not a lot of time available for it ;-)

No worries too.
So in case of : 

Here are a few ASM test example : 
https://github.com/SpinalHDL/VexRiscv/tree/master/src/test/cpp/custom

The SIMD_ADD  custom instruction test is probably a good template : 
https://github.com/SpinalHDL/VexRiscv/blob/master/src/test/cpp/custom/simd_add/src/crt.S

Then you can add the test in the regression the same way than : 
https://github.com/SpinalHDL/VexRiscv/blob/master/src/test/cpp/regression/main.cpp#L3861

More generaly, you can follow the way how SIMD_ADD was added (see the bottom of the chapter for how to run it) : 
https://github.com/SpinalHDL/VexRiscv#add-a-custom-instruction-to-the-cpu-via-the-plugin-system

Good luck :)

Dolu1990:
I'm currently working on some AES custom instruction, and i kind of have a similar case with permutation / rotate.

It isn't pushed yet, but there is the "interresting" part where rotation / permutation are done : 
(Basicaly, it works by bulding the index values, and only at the very end it does the required data mux)

```scala
    memory plug new Area{
      import memory._
      val romData = rom.readSync(onExecute.romAddress, !arbitration.isStuck)
      val romBytes = romData.subdivideIn(8 bits)

      val romSel = new Area {
        val output = Vec(UInt(2 bits), 4)
        def assign(l: Int*) = (output, l).zipped.foreach(_ := _)

        when(!input(INSTRUCTION)(mapping.DECRYPT)) {
          when(!input(INSTRUCTION)(mapping.LAST)) {
            assign(0, 1, 1, 2)
          } otherwise {
            assign(1, 1, 1, 1)
          }
        } otherwise {
          when(!input(INSTRUCTION)(mapping.LAST)) {
            assign(0, 1, 2, 3)
          } otherwise {
            assign(3, 3, 3, 3)
          }
        }
      }

      val rotate = new Area{
        val zero = B"0000"
        val byteSel = input(INSTRUCTION)(mapping.SEL, 2 bits).asUInt
        val output = Vec(Bits(8 bits), 4)

        def romSelMap(l : Int*) = Vec(l.map(romSel.output(_)))
        val sel = byteSel.mux(
          3 -> romSelMap(0, 1, 2, 3),
          2 -> romSelMap(1, 2, 3, 0),
          1 -> romSelMap(2, 3, 0, 1),
          0 -> romSelMap(3, 0, 1, 2)
        )
        when(input(INSTRUCTION)(mapping.LAST)){
          zero := B"1111"
          zero(byteSel) := False
        }

        for(byteId <- 0 to 3){
          output(byteId) := romBytes(sel(byteId))
          when(zero(byteId)){
            output(byteId) := 0
          }
        }
      }

      val xored = rotate.output.asBits ^ input(RS1)

      insert(CALC) := xored
    }
``` 

I post this message for your interrest, you realy don't need to update the PR with the above way of doing things.
First having working thing is the most important, then, maybe optimisation if time and enjoyement allow it ^^

rdolbeau:
Thanks for the example, not sure I understand it... I'll try harder :-)

Any idea why this semantic doesn't seem to work:

```
S	MIN	"((input(SRC1).asSInt < input(SRC2).asSInt) ? input(SRC1) | input(SRC2))"
S	MAX	"((input(SRC1).asSInt > input(SRC2).asSInt) ? input(SRC1) | input(SRC2))"
S	MINU	"((input(SRC1).asUInt < input(SRC2).asUInt) ? input(SRC1) | input(SRC2))"
S	MAXU	"((input(SRC1).asUInt > input(SRC2).asUInt) ? input(SRC1) | input(SRC2))"
```

The weird syntax is that I've given up on copy/pasting, I'm mostly auto-generating the plugin now, so the stuff in double-quotes is what should be used to implement the function in the second field. This goes all the way into supporting the instructions in the core/soc just fine, but the results are often wrong. (I _think_ it's the semantic, as every other instructions seem to work in the infrastructure and they are all tested the same way from the C/intrinsics code and the patterns are directly extracted from the documentation... though it could be a discrepancy between the compiler and doc...).

```
_rv32_min(0x00000001, 0x00000002) -> 0x00000001 -- OK
_rv32_minu(0x00000001, 0x00000002) -> 0x00000002 -- NOK
_rv32_max(0x00000001, 0x00000002) -> 0x00000001 -- NOK
_rv32_maxu(0x00000001, 0x00000002) -> 0x00000002 -- OK
```
And it's not that they always return SRC1 or SRC2:
```
_rv32_min(0x00000002, 0x00000001) -> 0x00000001 -- OK
_rv32_minu(0x00000002, 0x00000001) -> 0x00000002 -- NOK
_rv32_max(0x00000002, 0x00000001) -> 0x00000001 -- NOK
_rv32_maxu(0x00000002, 0x00000001) -> 0x00000002 -- OK
```
And it's not always the same functions that are broken either:
```
_rv32_min(0xffffffff, 0x00000001) -> 0xffffffff -- OK -1 < 1
_rv32_minu(0xffffffff, 0x00000001) -> 0x00000001 -- OK 1 < 2^32-1
_rv32_max(0xffffffff, 0x00000001) -> 0x00000001 -- OK 1 > -1
_rv32_maxu(0xffffffff, 0x00000001) -> 0xffffffff -- OK 2^32-1 > 1
```
I'm at a loss there. Indeed I must have been an heretic thinking SpinalHDL was somewhat easy; it's really hard actually :-)

rdolbeau:
@Dolu1990 As not everyone will want all instructions, I've made a quick'n'dirty code generator for this:

https://github.com/rdolbeau/VexRiscvBPluginGenerator

It generates a plugin similar to this one with a two-level mux for a user-specified set of B instructions. They are not all implemented, as I'm not sure on how to implement some of them, and I'm not sure the three-operands are doable. There's also a C test file to stress the instructions in the generated hardware.

There's probably room for tuning as it takes quite a bit of FPGA space to add everything:

```
Usage in Artix-7 35T for the whole LiteX SoC without B:
+----------------------------+------+-------+-----------+-------+
|          Site Type         | Used | Fixed | Available | Util% |
+----------------------------+------+-------+-----------+-------+
| Slice LUTs                 | 5400 |     0 |     20800 | 25.96 |
|   LUT as Logic             | 5230 |     0 |     20800 | 25.14 |
|   LUT as Memory            |  170 |     0 |      9600 |  1.77 |
|     LUT as Distributed RAM |  168 |     0 |           |       |
|     LUT as Shift Register  |    2 |     0 |           |       |
| Slice Registers            | 4660 |     0 |     41600 | 11.20 |
|   Register as Flip Flop    | 4660 |     0 |     41600 | 11.20 |
|   Register as Latch        |    0 |     0 |     41600 |  0.00 |
| F7 Muxes                   |   78 |     0 |     16300 |  0.48 |
| F8 Muxes                   |   13 |     0 |      8150 |  0.16 |
+----------------------------+------+-------+-----------+-------+
Usage in Artix-7 35T for the whole LiteX SoC with 44 B instructions (all implemented now):
+----------------------------+------+-------+-----------+-------+
|          Site Type         | Used | Fixed | Available | Util% |
+----------------------------+------+-------+-----------+-------+
| Slice LUTs                 | 6731 |     0 |     20800 | 32.36 |
|   LUT as Logic             | 6561 |     0 |     20800 | 31.54 |
|   LUT as Memory            |  170 |     0 |      9600 |  1.77 |
|     LUT as Distributed RAM |  168 |     0 |           |       |
|     LUT as Shift Register  |    2 |     0 |           |       |
| Slice Registers            | 4674 |     0 |     41600 | 11.24 |
|   Register as Flip Flop    | 4674 |     0 |     41600 | 11.24 |
|   Register as Latch        |    0 |     0 |     41600 |  0.00 |
| F7 Muxes                   |  134 |     0 |     16300 |  0.82 |
| F8 Muxes                   |   13 |     0 |      8150 |  0.16 |
+----------------------------+------+-------+-----------+-------+
```

Dolu1990:
> _rv32_minu(0x00000001, 0x00000002) -> 0x00000002 -- NOK

That's weird. Your macro seem ok. Can you send me the verilog i can just check it in case of XD

> Unless I have my binary very wrong, 101 is 5 and 110 is 6 and that's not coherent. Yeah, SpinalHDL is back to being easy ;-)

Sorry, that was the issue ?

> I'm not sure the three-operands are doabl

Currently that's not supported, should need a update in the RegFilePlugin and the HazardPlugin

> There's probably room for tuning as it takes quite a bit of FPGA space to add everything:

Welcome in hell ^^

rdolbeau:
> Sorry, that was the issue ?

Software problem, the toolchain isn't up to date with the specifications... fixing the toolchain fixed the issue (I only discovered the problem after asking the question, because Murphy's law)

> Currently that's not supported, should need a update in the RegFilePlugin and the HazardPlugin

OK, that's what I figured. There's a lot of good 3-operands instruction semantic out there, but I understand it makes things more complicated in the hardware...

> Welcome in hell ^^

Something like that :-)

lindemer:
@rdolbeau Would you mind sharing how you generated those FPGA hardware usage statistics for VexRiscv/LiteX? I'd like to do the same for a project I'm working on.

rdolbeau:
@lindemer They're just the output that Vivado produces in the various 'top_utilization*.rpt' files, specifically 'top_utilization_place.rpt' if I remember correctly:

```
[dolbeau@localhost linux-on-litex-vexriscv]$ ll build/sbustoztex/gateware/top_utilization_place.rpt 
-rw-rw-r--. 1 dolbeau dolbeau 10968 Jan 13 03:30 build/sbustoztex/gateware/top_utilization_place.rpt
```

lindemer:
Thanks!

rdolbeau:
I'll close that one, the plugin generator has grown (https://github.com/rdolbeau/VexRiscvBPluginGenerator/) and there's little point in adding some specific subset of B.

Dolu1990:
Ok :)
I'm also working on a FPU for vex, which has the specificity that the FPU can be shared between multiple CPU. Baybe that scould be a thing to share hardware for AES acceleration and heavy things like this.

rdolbeau:
@Dolu1990 Oh nice, though I suspect even one FPU won't fit in smaller FPGA... But I won't be complaining about the idea of sharing the FPU between cores :-)

For B there's so many subsets and combinations, any idea how it could be optionally integrated in VexRiscV ? (when it's standardized, that is).

For AES I would suggest investigating the move from custom to Zkne when it's standardized as your implementation seems very close/identical in semantic; my (software-like ;-) ) version is currently stuck at 0.8.0 of the Kscalar extension, as 0.8.1 is sourcing the destination register from the field rs1 instead of the field rd, and I'm not sure how to 'fix' VexRiscV for that... it should be possible to temporarily change 'rdRange' for those specific instructions, but it feels very hackish... that's probably going to be the big problem to implement Zkne in small designs.

Dolu1990:
> For B there's so many subsets and combinations, any idea how it could be optionally integrated in VexRiscV ? (when it's standardized, that is).

I don't know much about the B extension. So it's hard to me to have a good view of the situation.
Maybe the whole B extension should be implemented into a single plugin who is smart enough to share hardware inside itself.

Else in VexRiscv itself, there is so form of sharing, for instance the SrcPlugin is probably one of the most shared piece of hardware which provide two source operands muxes, and also add/sub. Then other plugin add to the DecoderPlugin the proper decoding spec to drive the instruction into the SrcPlugin properly for their usage.

>  as 0.8.1 is sourcing the destination register from the field rs1

wtf XD
Hmm in worst case that can be an option in the hazard plugin and the register file plugin to handle those corner cases, but that's realy hakish as you said.

rdolbeau:
> > as 0.8.1 is sourcing the destination register from the field rs1
> wtf XD

Exactly :-)

> Hmm in worst case that can be an option in the hazard plugin and the register file plugin to handle those corner cases, but that's realy hakish as you said.

That was all that was needed; I have a working patch on https://github.com/rdolbeau/VexRiscv/tree/three_operands , matching plugin/AES code in https://github.com/rdolbeau/VexRiscvBPluginGenerator/ . Works-For-Me(tm)...

PR #158, opened 2020-12-15
Fix DBusSimplePlugin catchAccessFault for store operations.

losfair:
Currently the logic in `DBusSimplePlugin` only raises access faults on loads from DBus, but not stores.

This PR extends access fault checking to stores.

### DIFF
--- a/src/main/scala/vexriscv/plugin/DBusSimplePlugin.scala
+++ b/src/main/scala/vexriscv/plugin/DBusSimplePlugin.scala
@@ -488,9 +488,9 @@ class DBusSimplePlugin(catchAddressMisaligned : Boolean = false,
         memoryExceptionPort.code.assignDontCare()
         memoryExceptionPort.badAddr := input(REGFILE_WRITE_DATA).asUInt
 
-        if(catchAccessFault) when(dBus.rsp.ready && dBus.rsp.error && !input(MEMORY_STORE)) {
+        if(catchAccessFault) when(dBus.rsp.ready && dBus.rsp.error) {
+          memoryExceptionPort.code := (input(MEMORY_STORE) ? U(7) | U(5)).resized
           memoryExceptionPort.valid := True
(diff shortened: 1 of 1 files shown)


losfair:
There's a test failure at https://travis-ci.org/github/SpinalHDL/VexRiscv/jobs/749769389#L2950 . The corresponding code is https://github.com/SpinalHDL/VexRiscv/blob/master/src/test/cpp/raw/machineCsr/src/crt.S#L141 .

It seems that the reference trace assumes that storing to a bad address (`0xF00FFF60`) doesn't fault, but it does now. And the new behavior looks correct to me.

I'm not familiar with how the reference trace is generated - is it from a previous version of VexRiscv, or something like a co-simulation model?

Dolu1990:
Hi, 

Ahhhhh i got it
Basicaly, VexRiscv DBusSimplePlugin do not want a response on writes, so it can't generate a store fault exception.
bus read -> bus response -> check error and may trap
bus write -> nothing -> can't trap

Did that to optimise store.

The same is true for DBusCachedPlugin, as it is write through, it do not want to wait for each write response (which can go through a long 30 cycle delay for a DDR) before continuing the programm execution.

make sense ?

Thanks anyway :)

Dolu1990:
> I'm not familiar with how the reference trace is generated - is it from a previous version of VexRiscv, or something like a co-simulation model?

Basicaly, there is a software golden model in main.cpp which is cross checked with VexRiscv hardware simulation. Both need to be updated together ^^

But in the given case, as the patch fix a "wanted feature", i think we can close it.

losfair:
Yeah got it. From the name `catchAccessFault` I assumed that it catches both load and store errors, and I just found that the [documentation](https://github.com/SpinalHDL/VexRiscv#dbussimpleplugin) describes it as "If a memory **read** returns an error, it will fire a trap exception".

Thanks!

PR #161, opened 2021-01-13
Distinguish between page faults from the MMU and access faults from PMP

lindemer:
The PMP plugin currently emits instruction, load and store **page** faults (12, 13, 15) but it should emit **access** faults (1, 5, 7). After digging into the code, I realized all of the existing memory translator bus references assume an MMU, not PMP, so I renamed a lot of variables to be more general. All of this is a prerequisite for a combined MMU + PMP plugin, which I may get around to later.

I also renamed the the instruction "catch" variables so they match their meaning in the RISC-V specification. This should make future changes a bit easier to understand:
- `catchIllegalAccess` is now `catchInstructionPage` (12)
- `catchAccessFault` is now `catchInstructionAccess` (1)
- `catchAccessError` is now `catchLoadStoreAccess` (5, 7)
- `catchIllegal` is now `catchLoadStorePage` (13, 15)
- `catchUnaligned` is now `catchLoadStoreMisaligned` (4, 6)

### DIFF
--- a/src/main/scala/vexriscv/Services.scala
+++ b/src/main/scala/vexriscv/Services.scala
@@ -71,6 +71,7 @@ case class MemoryTranslatorCmd() extends Bundle{
 case class MemoryTranslatorRsp() extends Bundle{
   val physicalAddress = UInt(32 bits)
   val isIoAccess = Bool
+  val isPaging = Bool
   val allowRead, allowWrite, allowExecute = Bool
   val exception = Bool
   val refilling = Bool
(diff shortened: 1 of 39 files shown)


Dolu1990:
Hi,

Hoo quite a few changes XD

One thing is PR which aren't urgent fixes should be on the dev branch, not master.
How is it to rebase it on dev ?

One issue i have, is that the PR is "big" with mostly refractoring of variable names, idealy, refractoring should have it's own dedicated PR, to avoid me missing bad non refractoring change lost in the middle XD

What is val isPaging = Bool about ?

Differeciate between MMU and PMP ?

lindemer:
That makes sense 😅 Okay I'll resubmit this as a PR to dev for the renaming and a PR to master for the page fault fix. I'll explain the fix in the new PR.

PR #162, opened 2021-01-14
Distinguish between page faults from MMU and access faults from PMP

lindemer:
VexRiscv was originally implemented before the introduction of PMP, so the `MemoryTranslatorRsp` class has no way of differentiating between faults resulting from the MMU and faults resulting from PMP. This PR adds a new flag called `isPaging` to indicate whether the `allowRead`, `allowWrite` and `allowExecute` flags refer to MMU or PMP permissions. This flag is handled by the data and instruction bus plugins to determine whether to throw access faults or page faults.

This fixes a bug where the PMP plugin throws page faults (12, 13, 15) instead of access faults (1, 5, 7) as the specification expects.


Dolu1990:
Nice thanks :D

PR #167, opened 2021-03-24
Attempt at supporting C (ompressed) and F/D (floating-point) together

rdolbeau:
Add support for the 8 C/FP L/S instructions. Enable a RV32IMAFDC buildroot to boot further than before, but not to reach login.

To be validated before merging...


Dolu1990:
The PR match what i had localy :D
Thanks :)

PR #174, opened 2021-04-09
New PMP plugin optimized for FPGAs

lindemer:
This PR restructures the PMP plugin, which significantly reduces LUT and maximum delay path on FPGAs. On Arty A7 35T with 16 PMP registers, the LUT utilization is reduced from 40.28% to 31.45% (from a baseline of 23.62%). This was achieved with the following modifications:

- The PMP CSR decoder logic is now centralized (before, it was duplicated for each register).
- Writes to PMP configuration registers are now done sequentially (2-3 cycles for `pmpaddr#`, 5 cycles for `pmpcfg#`), which is up from 2 cycles for other CSRs.
- All of the encoded `pmpaddr#` registers are stored in distributed RAM.


Dolu1990:
Hi,

Cool, that quite a area reduction XD

A few things before merges : 

1) Feature, especialy if they change vital things (as the CsrPlugin), should be done in the dev branch for two reasons. First there is a automatic regression which run daily (not in travis) which generation 2000 random configuration of the CPU and test them. Second, i don't want to break anything in master.

2) This PR actualy introduce PMP stuff into the CsrPlugin, which shouldn't be the case idealy. The goal would be to extends https://github.com/SpinalHDL/VexRiscv/blob/dev/src/main/scala/vexriscv/plugin/CsrPlugin.scala#L360 to add require feature which would allow implementing a self contained PMP. Would that be possible ?

If there is implemenations "axiom" in the CPU that you miss, let's me know :)

lindemer:
@Dolu1990 I don't think it makes sense to extend `CsrInterface` for PMP because these registers are handled quite differently from any other CSRs. The changes I've made to `CsrPlugin` can basically be summed up as: _If a PMP CSR is accessed, do nothing._ Here's a breakdown of what happens when you write the PMP registers (by CPU cycle):

**pmpcfg(N)**

1. Perform a masked write based on which L-bits in the register were already set. If the new setting is TOR and locked, then the previous region also gets locked. 
2. Compute the bounds for region N * 4 + 0.
3. Compute the bounds for region N * 4 + 1.
4. Compute the bounds for region N * 4 + 2.
5. Compute the bounds for region N * 4 + 3.

**pmpaddr(N)**

1. Check if the L-bit is set on region N. If not, then write`pmpaddr(N)`, which is stored in asynchronous (distributed) RAM. 
2. Compute the bounds for region N.
3. (If N =/= regions - 1.) Compute the bounds for region N + 1. This is done in case region N + 1 uses TOR. 

The dependency between these two register types and the lock-masking logic makes it quite tricky to use the existing `CsrInterface`, and at the same time, I don't really see how this functionality could be added to it without adding a lot more PMP-specific code to `CsrPlugin`.

Dolu1990:
> If a PMP CSR is accessed, do nothing

Then we can add in the CsrInterface a primitive to say :
"Do not generate unimplemented CSR exception for this CSR address"
 ?

I would realy prefer having it in a neutral way than having it in a  hardcorded way.

lindemer:
Aha yes that would be much better, I agree. In that case I wouldn't need to modify anything in `CsrPlugin`. I'll see if I can do it that way; might come back with questions.

lindemer:
@Dolu1990 I think this last commit does what you had in mind. However, I still need to make this line conditional on whether the CSR being accessed has been mapped to `CsrIgnoreIllegal`: https://github.com/SpinalHDL/VexRiscv/blob/79bc09e69aa9b83dd4499b70ae2c4552edf9bcb9/src/main/scala/vexriscv/plugin/CsrPlugin.scala#L1128

Any thoughts on how to do that?

Dolu1990:
Hmmm i think first we should take a look at the "big picture".

in the PmpPlugin, do you handle the following cases : 
- Not having side effect if a instruction in the execute stage write the pmp csr, but is canceled by a jump/exception futher in the pipeline
- In the case the register file is readed in the execute stage, should not use the RS1 value until hazard are cleared.

It seems to me that those two cases aren't covered.

(that's one of the reasons why i'm realy trying to cover all CSR implementations via a proper CsrInterface API)

Maybe we should just provide a lower level API in the CsrInterface to give more controle to other plugins ?

lindemer:
I agree that a low-level API for `CsrInterface` would be a lot cleaner, but it will take me some time to get it done. I suppose we could have an API for:

- CSR read/write to async memory objects
- CSR read/write with special locking bits

Dolu1990:
> CSR read/write to async memory objects
CSR read/write with special locking bits

Those two seems too high level / specific for me

What's about a something in that kind : 
- readLowLevel[T<:Data](data : T, request : Bool,  hazard : Bool, decode : UInt => Bool) 
data is the data that will be given to the CsrPlugin ALU
request is assigned by the CsrPlugin to notify the necessity of producing a readable data
hazard is assigned by the custom plugin to hold on
decode is a landa function used to filter which CSR id we are interrested in

- writeLowLevel[T<:Data](data : T, request : Bool,  hazard : Bool, decode : UInt => Bool) 
data is the result of the CsrPlugin ALU that the custom plugin need to manage.

Would that be enough ? It give to the custom plugin all the tools to implement FSM behind the CSR management, (as far as i can see)

lindemer:
@Dolu1990 I had to take a break from this but I'm back now.

Could you explain the arguments to these functions a bit more for me? Not sure I understand exactly which are inputs or outputs. I have a limited amount of time left to spend on this (about 1 month), so would you be interested in implementing this API yourself? Otherwise I can't guarantee that I'll be able to get PMP done and upstreamed.

Dolu1990:
> I had to take a break from this but I'm back now

No worries ^^ 

> so would you be interested in implementing this API yourself?

Sure, i will give a try tomorrow :)

lindemer:
Thanks so much dude, I really appreciate it!

Dolu1990:
So i looked at readLowLevel/writeLowLevel and implemented it, but this still seems to high level to me XD

So, here is a alternative proposal (which mix API and direct accesses) :
- csr.duringWrite(body : => Unit) // execute body when a CSR write instruction is in execute stage, allow you to enable a FSM
- csr.durringRead(body : => Unit) // execute body when a CSR write instruction is in execute stage, allow you to enable a FSM
- execute.arbitration.haltItself := True // while your statemachine isn't done
- execute.input(instruction)(csrRange) // Get the CSR ID 
- csr.allowCurrentCsr() // Call it when the current CSR access is ok (no trap)
- csr.trapCurrentCsr() // Call it when the current CSR access is not ok (trap)
- csr.readData() : Bits // You can assign the returned bits to override what the CsrPlugin read
- csr.writeData() : Bits // Can be used to get the data that the CsrPlugin want to write

Would that make sense ? if you think one case isn't covered, tell me too ^^

lindemer:
Thanks :) I'm not so clear on what we would need `allowCurrentCsr()` and `trapCurrentCsr()` for. Here's another idea that could work but maybe it's too specific or complicated.

I would say that the two unique functions we need for PMP are:

1. Perform a masked CSR write based on some arbitrary locking bits.
2. **After the masked write is completed** (i.e., on the next cycle) start a state machine and continue to halt the pipeline until it's finished.
3. The same as (1.) except the writes will go to RAM instead of registers.

To accomplish (1.), we could add an API like this:
```
def rConditional(csrAddress : Int, bitOffset : Int, that : Data, condition : Bool): Unit
def wConditional(csrAddress : Int, bitOffset : Int, that : Data, condition : Bool): Unit
def rwConditional(csrAddress : Int, bitOffset : Int,that : Data, condition : Bool): Unit ={
  rConditional(csrAddress,bitOffset,that,condition)
  wConditional(csrAddress,bitOffset,that,condition)
}
def rwConditional(csrAddress : Int, thats : (Int, Data, Bool)*) : Unit = for(that <- thats) rwConditional(csrAddress,that._1, that._2, that._3)
```
Then, for the PMP configuration registers, I can declare them sort of like this:
```
csrService.rwConditional(0x3a0,
    31 -> pmp3.l -> pmp3.l, 27 -> pmp3.a -> pmp3.l, 26 -> pmp3.x -> pmp3.l, 25 -> pmp3.w -> pmp3.l, 24 -> pmp3.r -> pmp3.l,
    23 -> pmp2.l -> pmp2.l, 19 -> pmp2.a -> pmp2.l, 18 -> pmp2.x -> pmp2.l, 17 -> pmp2.w -> pmp2.l, 16 -> pmp2.r -> pmp2.l,
...)
```
To accomplish (2.), we could do the following. Since the state machine would start up the cycle _after_ the masked write completes, it can use the new values directly. This would require using `execute.input(instruction)(csrRange)` to configure the FSM.
```
csrService.onWrite(0x3a0) {
    execute.arbitration.haltItself := True
    // Start the FSM, which continues to halt the CPU until it's done.
}
```
I'm not sure how to do (3.) cleanly, but we could just start with this. Based on my research, adding the FSM had a much more significant performance impact than moving the `pmpaddr#` CSRs to RAM, anyways. What do you think?

**Also** I plan on removing the TOR and NA4 addressing modes, and making the granularity of the regions configurable in the top-level config. The reasons for this:

1. Most RTOSes only use aligned regions anyways, because that's what Arm, ARC, etc. have been doing for years. So, TOR won't even be used in most use cases.
2. The locking and decoding logic gets **a lot** more complicated with TOR enabled.
3. NAPOT region bounds can be computed with more efficient hardware logic. I haven't added it in this PR yet but I've already implemented it.

Dolu1990:
I pushed the new features to allow implementing the CSR in very custom ways, see : 
https://github.com/SpinalHDL/VexRiscv/commit/6066d8bc26a8dff72dd94a698575150fff4c850b#diff-6775b04f81e5c5960d1b10e4623c80c1617e26e6554d4cb7e4157fc89a5edebeR389
https://github.com/SpinalHDL/VexRiscv/commit/6066d8bc26a8dff72dd94a698575150fff4c850b#diff-6775b04f81e5c5960d1b10e4623c80c1617e26e6554d4cb7e4157fc89a5edebeR418


> Perform a masked CSR write based on some arbitrary locking bits.
> After the masked write is completed (i.e., on the next cycle) start a state machine and continue to halt the pipeline until it's finished.

I would say you can join both ?

```scala
val triggered = False
durringWrite{
   when(execute.input(INSTRUCTION)(csrRange) === myCsrId){ //You can also trig on a range of CSR and add more conditions
      triggered := True
   }
}

when(triggered){
      halt := True
      when(csrPlugin.isHazardFree()){
         //Your FSM her, using csr.writeData() to know what to write
      }
}
```

I mean, i guess that witht the new api provided, you can compose what you want ?

Let's me know if you have any issue ^^

> csrService.rwConditional(0x3a0,
    31 -> pmp3.l -> pmp3.l, 27 -> pmp3.a -> pmp3.l, 26 -> pmp3.x -> pmp3.l, 25 -> pmp3.w -> pmp3.l, 24 -> pmp3.r -> pmp3.l,
    23 -> pmp2.l -> pmp2.l, 19 -> pmp2.a -> pmp2.l, 18 -> pmp2.x -> pmp2.l, 17 -> pmp2.w -> pmp2.l, 16 -> pmp2.r -> pmp2.l,
...)

Why not putting the pmpX into a Vec ? and then using a for loop to map each of them one by one ? (instead of doing things by hand, which can't be precisly parametred) ex : i want only 2 pmp regions

> Also I plan on removing the TOR and NA4 addressing modes,

Removing some fat is always good ^^

lindemer:
Awesome, I see what you mean now :) I'll refactor my code and see if I can get it working.

lindemer:
Not sure if I'm using the `csrService.readData()` API correctly. CSR writes appear to work, but I can't read the values back. Am I doing something wrong here?
```
val enable = RegInit(False)
for (i <- 0 until regions) {
  csrService.onRead(0x3b0 + i) { csrService.readData().assignFromBits(cfgRegister(pmpSelect)) }
  csrService.onWrite(0x3b0 + i) { enable := True }
}
for (i <- 0 until (regions / 4)) {
  csrService.onRead(0x3a0 + i) { csrService.readData() := pmpaddr.readAsync(pmpIndex).asBits }
  csrService.onWrite(0x3a0 + i) { enable := True }
}
```

lindemer:
Do you mean like this?
```
val enable = RegInit(False)

csrService.duringAnyRead {
  when (input(PMPCFG)) { csrService.readData().assignFromBits(cfgRegister(pmpSelect)) }
  when (input(PMPADDR)) { csrService.readData() := pmpaddr.readAsync(pmpIndex).asBits }
}

csrService.duringAnyWrite {
  when (input(PMPCFG) | input(PMPADDR)) {
    csrService.allowCsr()
    enable := True
  }
}
```
So, in my last post, the CSR writes didn't cause an exception because any CSR marked with `CsrOnWrite` or `CsrOnRead` disables the `illegalAccess` flag. I guess `csrService.allowCsr()` should do the same thing here, but I get an exception anyways.

lindemer:
I changed the API a bit so that I can declare individual CSRs as valid. I think this makes more sense than calling `csrService.allowCsr()` when a match occurs. 
```
for (i <- 0 until regions) csrService.allow(0x3b0 + i)
for (i <- 0 until (regions / 4)) csrService.allow(0x3a0 + i)
```
I still can't figure out how to get the `csrService.readData()` API to work, though. It looked correct in my last post but it doesn't return the correct values.

Dolu1990:
> I guess csrService.allowCsr() should do the same thing here, but I get an exception anyways.

Right it should. hmmm looking at the CsrPlugin it should works (illegalAccess clearWhen(csrMapping.allowCsrSignal))


> csrService.allow(0x3b0 + i)

Basicaly, i wanted to avoid that as it is less flexible.

> input(PMPCFG)  input(PMPADDR)

Whare are those exactly ? maybe the issue come from there ?

> csrService.duringAnyRead {

maybe, try to ensure it is properly executed in the simulation. by setting a summy signal to true when executed. Just for sanity XD

lindemer:
Those are some stageables I added to check if the CSR access is for PMP:
```
decode plug new Area {
  import decode._
  insert(PMPCFG) := input(INSTRUCTION)(31 downto 24) === 0x3a
  insert(PMPADDR) := input(INSTRUCTION)(31 downto 24) === 0x3b
}
```
Looking at the VCD, I can see that the CSR writes and the FSM are still working correctly. When the Assembly test tries to do a `csrr` on any of the PMP registers, it doesn't match the actual content. It's a problem for both `pmpcfg` (registers) and `pmpaddr` (RAM), so the source appears to be the `csrService.readData()` API.

Dolu1990:
I just pushed a fix in the CsrPlugin.CsrMapping. That way, now , it give a names to the internal signals.

So, normaly, you have the following signals used by the CsrPlugin to get the read value from the CSR : 

  wire       [31:0]   CsrPlugin_csrMapping_readDataInit;
  wire       [31:0]   CsrPlugin_csrMapping_readDataSignal;
  
  CsrPlugin_csrMapping_readDataInit is generated by the original CSR API.
  
  While CsrPlugin_csrMapping_readDataSignal is by default assigned by CsrPlugin_csrMapping_readDataInit, then accessible via csr.readData(), and then used by the CsrPlugin as final readed value.
  
  In a first time, maybe try to not overcomplicated, by this i mean : 
  
  ```scala
1) 
csrService.readData() := 0x12345678

2)
csrService.duringAnyRead {
     csrService.readData() := 0x12345678
}
  
3)
 csrService.duringAnyRead {
  when (input(PMPCFG)) { csrService.readData() := 0x12345678 }
  when (input(PMPADDR)) { csrService.readData() := 0x12345678 }
}
  
4)
 csrService.duringAnyRead {
  when (input(PMPCFG)) { csrService.readData().assignFromBits(cfgRegister(pmpSelect)) }
  when (input(PMPADDR)) { csrService.readData() := pmpaddr.readAsync(pmpIndex).asBits }
}
```
  
In other words, trying first to get the very minimal to works, and incrementaly add layers on the top, that way, sanity is preserved <3

Let's me know how things goes :)

**Comment by lindemer** [comment]
@Dolu1990 Almost finished :) I have it working with `duringAnyRead` now. For the write operation, I think I need an `onAnyWrite` API, not `duringAnyWrite`. The `during` signals are on for the duration of the operation, and the `on` signals are only toggled for one cycle at the beginning, right?

**Comment by lindemer** [comment]
@Dolu1990 Think we're good now. Lemme know if you have questions 🚀

**Comment by Dolu1990** [comment]
> @Dolu1990 Think we're good now. Lemme know if you have questions rocket

Cool :D

One thing to add is isHazardFree

Basicaly, a CSR instruction can reach isHazardFree while there is still instruction down (memory/writeback stage) which could unschedule the CSR instruction.
There is also the case where the registerfile plugin is configured to be readed in the execute stage itself, in such case the registerfile parameters of the CSR instruction may not be right in a first time.

To avoid this, you should not have any side effect, nor start your stage machine until csr.isHazardFree is set.

Does that make sense ?

**Comment by lindemer** [comment]
Yep, I think this code does that already:
```
csrService.onAnyWrite {
  when ((pmpcfgCsr | pmpaddrCsr) & machineMode) {
    pending := True
  }
}

when (pending) {
  arbitration.haltItself := True
  when (hazardFree & pmpaddrCsr_) {
    // write pmpcfg
  }
}

pmpaddr.write(pmpNcfg_, writeData_.asUInt, ~locked & pmpcfgCsr_ & pending & hazardFree)

val stateIdle : State = new State with EntryPoint {
  onExit {
    enable := True
    pending := False
    arbitration.haltItself := True
  }
  whenIsActive {
    when (pending & hazardFree) {
      when (pmpaddrCsr_) {
        goto(stateCfg)
      }.elsewhen (pmpcfgCsr_) {
        goto(stateAddr)
      }
    }
  }
}
```
So basically, the `pending` flag is set when a write is in the execute stage, but then it just halts until `hazardFree` is `True`. The CSR write and FSM don't do anything until the hazards are clear.

**Comment by Dolu1990** [comment]
> Yep, I think this code does that already:

Greate ^^
Yes i can see that now.

> Edit Do I need to do something if the instruction gets flushed? Like maybe I should set pending back to false? Something like:

No, basicaly if hazardFree is set, the CSR instruction will never be flushed and will fully execute.

**Comment by lindemer** [comment]
Right, I just added in this snippet because I think there could be issues otherwise:
```
when (arbitration.isFlushed) {
  pending := False
}
```
Like, if a jump occurs, the `pending` flag should be toggled off, or else the CSR write will happen at some random time in the future even though the instruction got flushed (i.e., whenever `hazardFree` goes high). I didn't write a test for this but I'm pretty sure this is necessary.

**Comment by lindemer** [comment]
Otherwise, I think we're good to merge?

**Comment by Dolu1990** [comment]
Hmmm

I can see your statemachine kick of after the csr instruction. Right ?

Why not kicking that statemachine durring the csr instruction, and holding the instruction until the FSM is done ?

That way, all the side efect of the state machine are completeded before letting the CPU going futher ?

(https://github.com/SpinalHDL/VexRiscv/pull/174/files#diff-334d1ad66ec93f3dfdbbc19a5b1f2315e6a096aad250cdc57787bca469069e29R168)

**Comment by lindemer** [comment]
Basically the order of events is:

1. Set `pending` to `True`.
2. Once `hazardFree` is `True`. Perform the write, toggle `pending` to `False`, toggle `enable` to `True`.
3. The FSM starts updating the PMP region bounds the cycle _after_ the write, because it needs to use the _result_ of the write after the locking check. If it used the user-supplied CSR value, it would allow the user to bypass the lock mechanism.

Does that make sense? I think all the side effects are completed before the CPU is un-halted, but maybe I misunderstand the question.

**Comment by Dolu1990** [comment]
I'm a bit worring of corner cases,

For instance : 
 arbitration.haltItself := True   should only be used by the instruction in the given stage, and some instruction will make that assumption.  So in the case you have a csr to the pmp and then one instruction doing that assumption it will likely make things glitch.
 
 >  because it needs to use the result of the write after
 
 what if you implement the read and the write in the FSM while the CSR is stuck by a "haltItself" in the execute stage ? (that was what i was thinking about when putting the new API) ^^

**Comment by Dolu1990** [comment]
So basicaly, not using anymore the onWrite/onRead (which are single cycle pulse at the very end of the instruction)
But instead using the durringWrite/durringRead to implement things ?

**Comment by lindemer** [comment]
> what if you implement the read and the write in the FSM while the CSR is stuck by a "haltItself" in the execute stage ? (that was what i was thinking about when putting the new API) ^^

I separated the `writer` and `controller` areas for readability and because the `pmpaddr.write` command can't be included in the FSM, because it's an "always" action and doesn't depend on state. If I put them together, the `pmpaddr` and `pmpcfg` writes will be separate and I think it will be more confusing. These only happen when the CPU is halted, already.

> So in the case you have a csr to the pmp and then one instruction doing that assumption it will likely make things glitch.

Sorry can you explain this differently? I don't understand :/ I'm halting the CPU and waiting to see if the instruction will be flushed, which is exactly what `CsrPlugin` does, if I understand correctly.

> So basicaly, not using anymore the onWrite/onRead (which are single cycle pulse at the very end of the instruction)
But instead using the durringWrite/durringRead to implement things ?

`duringAnyRead` works for performing the reads, because these can happen even if the instruction gets flushed (if I understand correctly). The writes need to use `onAnyWrite` instead of `duringAnyWrite`, because I need to save the current instruction and write data to registers at precisely that moment before I do the write. When I was debugging this with the VCD, I noticed that those signals change throughout `duringAnyWrite`, so I can't use them directly to do the PMP reconfiguration.

**Comment by Dolu1990** [comment]
> because these can happen even if the instruction gets flushed (if I understand correctly)

If isHazardFree is true =>  the instruction will never be flushed :)

> Sorry can you explain this differently? I don't understand :/ I'm halting the CPU and waiting to see if the instruction will be flushed, which is exactly what CsrPlugin does, if I understand correctly.

There is haltItself and haltByOthers. For instance a store instruction will not put a transaction on the bus until haltByOthers is cleared, then the store will haltItself until the bus transaction moved out of the CPU. In other words. the behaviour of some instruction will change depending which one is set.

> because I need to save the current instruction and write data to registers at precisely that moment before I do the write

Right for the registers. For the ram, i would say the best would be to readSync them (for timing reasons) which will also buffer the value, allowing then to write that same ram before the end of the csr instruction in the execute stage.

Right ?

(I'm not saying that the actual way do not works)

**Comment by lindemer** [comment]
Okay I think we might be describing the same thing, but I'm not sure. The process that I've implemented is like this:
```
(idle) -- onWrite  --> (pending, haltItself) -- hazardFree --> (enable) -- run the FSM --> (idle)
                           |
                           +------- isFlushed ----> (idle)
```
I'm checking whether the instruction has been flushed _before_ `hazardFree` is set to `True`. If it gets flushed, then it goes back to its default state and stops waiting for the `hazardFree` signal.

Could you give me a specific list of changes/tests you want before the merge? Really appreciate the feedback, I'm just in a bit of a rush 😄

**Comment by lindemer** [comment]
A call would be great, I think I'm missing the nuances of how the CSR plugin works 😅 Do you have any time this afternoon (CET)?

**Comment by lindemer** [comment]
@Dolu1990 This fixes what we discussed except for the flush/jump after the write.

- `pending` is now cleared when `!execute.arbitration.isStuck`
- It only uses `duringAnyWrite` now, which ensures that the CSR write and PMP reconfiguration completes while the instruction is still in the execute stage. (Before it was actually halting the execute stage after the instruction had moved on.)
- Everything is in one FSM and I got rid of the `readAsync`s.

**Comment by Dolu1990** [comment]
Good thanks ^^

I guess there is one last thing.
Basicaly, if you have the following code sequance : 

```
j miaou
...
miaou:
   csrrw to the pmp
```

the CSR instruction will not be blocked in the execute stage, but go directly through.
The reason is that as the pipeline will clean and flushed by the previous jump, the CsrPlugin will not wait any cycles.
meanwhile, you set the pending register, but as this a one cycle delay until it set the haltItSelf it will be "too late".

So, i would say : 

```scala
      val fsmCompletion := False
      csrService.duringAnyWrite {
        when ((pmpcfgCsr | pmpaddrCsr) & machineMode) {
           arbitration.haltItself := !fsmCompletion

       val stateAddr : State = new State {
          onEntry (counter := pmpNcfg_) //! Be aware that the on entry occure when the fsm is in the last cycle of the previous state !
          whenIsActive (fsmCompletion := True; goto(stateIdle))
        }
        
```

Dolu1990:
About the 

> So, i would say : 
```scala
      val fsmCompletion := False
      csrService.duringAnyWrite {
        when ((pmpcfgCsr | pmpaddrCsr) & machineMode) {
           arbitration.haltItself := !fsmCompletion

       val stateAddr : State = new State {
          onEntry (counter := pmpNcfg_) //! Be aware that the on entry occure when the fsm is in the last cycle of the previous state !
          whenIsActive (fsmCompletion := True; goto(stateIdle))
        }
        
```

Those arbitration.haltItself stuff would remplace the ones in the state machine

lindemer:
@Dolu1990 Cool, makes sense. Fixed 😄

Dolu1990:
Thanks ^^
Then i guess things are ready to be merged right ?

lindemer:
Yep, all done 😎

Dolu1990:
thanks ^^

lindemer:
@Dolu1990 When do you plan on merging dev with master?

Dolu1990:
@lindemer Merges to master need to be in sync with SpinalHDL releases. So in general i do it when major feature / bug / need 

But the nightly regression of vexriscv failed (it run 2000 random configuration).
It seems related the PMP : 
https://file.io/EFpNF0rjTOSg

To open this kind of very big logs, you can use glogg

ctrl + f   FAILURE, then scroll up,

```
TEST ERROR CODE 0
FAIL pmp at PC=80000370 time=396
TEST ERROR CODE 0
FAIL pmp at PC=80000370 time=392
TEST ERROR CODE 0
FAIL pmp at PC=80000370 time=402
TEST ERROR CODE 0
FAIL pmp at PC=80000370 time=380
TEST ERROR CODE 0
FAIL pmp at PC=80000370 time=380
TEST ERROR CODE 0
FAIL pmp at PC=80000370 time=394
TEST ERROR CODE 0
FAIL pmp at PC=80000370 time=382
TEST ERROR CODE 0
FAIL pmp at PC=80000370 time=386
TEST ERROR CODE 0
FAIL pmp at PC=80000370 time=412
TEST ERROR CODE 0
FAIL pmp at PC=80000370 time=386
``` 

So to reproduce the issue on your side : 

``` 
export VEXRISCV_REGRESSION_CONFIG_COUNT=2000
export VEXRISCV_REGRESSION_CONFIG_LINUX_RATE=0.3
export VEXRISCV_REGRESSION_CONFIG_MACHINE_OS_RATE=0.5
export VEXRISCV_REGRESSION_FREERTOS_COUNT=2
export VEXRISCV_REGRESSION_ZEPHYR_COUNT=2
export VEXRISCV_REGRESSION_THREAD_COUNT=0
export VEXRISCV_REGRESSION_SEED=-3230758023236648713
export VEXRISCV_REGRESSION_TEST_ID=1969,1895,1867,1856,1826,1732,1727,1726,1662,1657,1651
sbt "testOnly vexriscv.TestIndividualFeatures" > sbt.log
glogg sbt.log
```

Then, you can find  the simulations in the simWorkspace folder with their test id.
If you want to rerun manualy a test with another set of arguements (for instance to have a proper wave), you can  cd into that test id folder, and do a make (with the proper arguements)

To get those proper agument look at the sbt.log. For instance to rerun test 1651 copy the make run REGRESSION_ .... whole line  : 
go to START TEST test_id_1651_ ....
...
[Done] at 2.368
make run REGRESSION_PATH=../../src/test/cpp/regression VEXRISCV_FILE=VexRiscv.v WITH_USER_IO=no REDO=10 TRACE=yes TRACE_START=100000000000ll FLOW_INFO=no STOP_ON_ERROR=no DHRYSTONE=yes COREMARK=yes THREAD_COUNT=1  SEED=1454466621 IBUS=CACHED IBUS_DATA_WIDTH=32 COMPRESSED=yes DBUS=CACHED DBUS_LOAD_DATA_WIDTH=32 DBUS_STORE_DATA_WIDTH=32 LRSC=yes AMO=yes  MUL=yes DIV=yes      CSR=yes CSR_SKIP_TEST=yes FREERTOS=0 ZEPHYR=2  DEBUG_PLUGIN=no MMU=no PMP=yes

Let's me know if you have any issues

Dolu1990:
typicaly, to get a wave, you can change REDO=1 TRACE_START=0 STOP_ON_ERROR=yes

lindemer:
When I click, the link, it says that the file has been deleted. Can you upload it again? Also, when I try to run the commands to reproduce the issue, I'm getting this error:
```
[info] Compiling 3 Scala sources to /home/sam/Shared/VexRiscv/target/scala-2.11/test-classes ...
[error] /home/sam/Shared/VexRiscv/src/test/scala/vexriscv/TestIndividualFeatures.scala:860:25: value doCmdWithLog is not a member of object spinal.lib.DoCmd
[error]         val str = DoCmd.doCmdWithLog(cmd, workspace)
[error]                         ^
[error] one error found
[error] (Test / compileIncremental) Compilation failed
```

Dolu1990:
> one error found 

I guess you need to update the SpinalHDL you have

There is is  :
[log (10).tar.gz](https://github.com/SpinalHDL/VexRiscv/files/6599212/log.10.tar.gz)

lindemer:
I went through the log, and it looks like the PMP test fails consistently with these settings:
```
new RegFilePlugin(
  readInExecute = true
)
new SrcPlugin(
  executeInsertion = true
)
```
Any idea why this would be the case? I built the CPU with these settings on my PC and ran the PMP test again, but it still passes. So, I haven't managed to replicate the failure.

Dolu1990:
Thanks ^^

Yes that's right.

I have a fix, basicaly the issue was that 

```scala
          when (!fsmPending) {
            fsmPending := True
            writeData_ := csrService.writeData()  //Here are the issue, it capture those state before they are hazard free (register file in execute stage can produce hazard on the data to write)
            pmpNcfg_ := pmpNcfg
            pmpcfgN_ := pmpcfgN
            pmpcfgCsr_ := pmpcfgCsr
            pmpaddrCsr_ := pmpaddrCsr
          }
```

I will push a fix soon, i just need to test more cases ^^

Dolu1990:
Fixed there : 
https://github.com/SpinalHDL/VexRiscv/commit/646911a37367cafc58f0e07a2d561ce395f26b13#diff-334d1ad66ec93f3dfdbbc19a5b1f2315e6a096aad250cdc57787bca469069e29L173
https://github.com/SpinalHDL/VexRiscv/commit/646911a37367cafc58f0e07a2d561ce395f26b13#diff-334d1ad66ec93f3dfdbbc19a5b1f2315e6a096aad250cdc57787bca469069e29L196

Other changes are there to allow to give a proper name to all elements in 
https://github.com/SpinalHDL/VexRiscv/commit/646911a37367cafc58f0e07a2d561ce395f26b13#diff-334d1ad66ec93f3dfdbbc19a5b1f2315e6a096aad250cdc57787bca469069e29R127

lindemer:
Awesome, thanks.

Dolu1990:
2000/2000 passed <3

PR #177, opened 2021-04-30
New Readme file to debug Murax SoC without usnig Jtag Adapter

Pradeep2004:


Readme
```diff
@@ -0,0 +1,74 @@
...
+2. Generation of Bscane2 within Murax SoC in Linux
+
+The BSCANE2 allows access between the internal FPGA logic and the JTAG Boundary Scan logic controller. This allows for communication between the internal running design and the dedicated JTAG pins of the FPGA.
+
+2.1 Steps to create Bscane2
+
+• After cloning all the files from https://github.com/SpinalHDL/VexRiscv, go to this path : src/main/scala/vexriscv/demo and find the Murax.scala file.
+
+• Comment out these lines to remove the toplevel jtag I/O pins:
+   Line 165, Line 395 to 397, Line 410 to 403
```
Dolu1990, line 16: It is dangerous to refer to lines, as the file may change over time.
I would say, either we need to put some little sections of code in the readme, either we need to refer to the code via a link which has a fixed commit. For instance : 

https://github.com/SpinalHDL/VexRiscv/blob/c489143442d9ccc23790153cca6ad33c2736114a/src/main/scala/vexriscv/demo/Murax.scala#L155

Dolu1990 (commented):
Thanks ^^

So i would say, the file can be moved into VexRiscv/doc/nativeJtag. Also you can place the .cfg files in it :)

Dolu1990:
Thanks ^^

PR #182, opened 2021-05-08
Make the [ID]TLB size configurable from Litex

rdolbeau:
This enables changing the # of [ID] TLB entries from Litex.

Matching PR in Litex is https://github.com/enjoy-digital/litex/pull/906


Dolu1990:
Thanks ^^

PR #184, opened 2021-05-17
fixed priority of == & != as seemed logical 

allexoll:
I has getting some 

```
../main.cpp:1198:24: warning: & has lower precedence than !=; != will be evaluated first [-Wparentheses]
            if(address & (size-1) != 0)
                       ^~~~~~~~~~~~~~~
```

when building, so i fixed the lines where there was a missing parenthesis. 

Remarks welcome


Dolu1990:
Thanks ^^

PR #220, opened 2021-11-11
Update DebugPlugin.scala: Add optional readback of hardware breakpoint values

BLangOS:
Add readback of the hardware breakpoint values.
A new parameter is added to the DebugPlugin to switch readback on and off.


Dolu1990:
Thanks :)

PR #233, opened 2022-01-26
plugin: CsrPlugin: Init cycle and instret registers

dnltz:
Both counters are initialized with "randBoot()". This is fine for FPGA
designs because the registers can be loaded with default values but
ASIC designs require to load the value during a reset.

Since both counters require to start at 0 (read-only CSR registers),
change both registers from "randBoot()" to "init(0)".

Error:

    reg        [63:0]   CsrPlugin_mcycle = 64'b0000000...00000000000;

           |
  Warning : Ignoring unsynthesizable construct. [VLOGPT-37]

Signed-off-by: Daniel Schultz <d.schultz@phytec.de>

### DIFF
--- a/src/main/scala/vexriscv/plugin/CsrPlugin.scala
+++ b/src/main/scala/vexriscv/plugin/CsrPlugin.scala
@@ -689,8 +689,8 @@ class CsrPlugin(val config: CsrPluginConfig) extends Plugin[VexRiscv] with Excep
         val exceptionCode = Reg(UInt(trapCodeWidth bits))
       }
       val mtval = Reg(UInt(xlen bits))
-      val mcycle   = Reg(UInt(64 bits)) randBoot()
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi ^^

So the idea of that randboot was not realy to specify some synthetisable behaviour, but only to initialize those register with a value in simulation to avoid X propagation.

But anyway, no worries having them with a reset value :)
Thanks

Dolu1990:
Hi @dnltz 

I think i have a good fix for the verilog randboot : 
https://github.com/SpinalHDL/SpinalHDL/commit/133a4e2cd400b4874c95e4bdc9219586575f205a

So, that way, you have a : 
  `ifndef SYNTHESIS
  
  to disable the randboot.

dnltz:
Hi @Dolu1990 - looks good :) Interesting ... such a basic feature was recently added to verilator.

PR #236, opened 2022-02-10
plugin: DBusSimplePlugin: Remove assert

dnltz:
This assert triggered sometimes at the beginning of a simulation.
Since it's not really needed anymore, we can remove it.

Signed-off-by: Daniel Schultz <daniel.schultz@aesc-silicon.de>


dnltz:
@Dolu1990 - ping :)

PR #243, opened 2022-04-02
Fix reset vector of GenCustomSimdAdd

andreasWallner:
With the old reset vector half of the tests fail since
they expect the CPU to start at 0x80000000.


Dolu1990:
Thanks ^^

PR #253, opened 2022-05-13
Update to latest risc-v-formal

mmicko:
This is just upstream change to support https://github.com/YosysHQ/riscv-formal/commit/891ea3694dfb4b47432f9fdaddff732005913283 that is now part of main riscv-formal release


Dolu1990:
Thanks :)

PR #288, opened 2022-12-20
Expand SATP register to 22 bits per spec

bunnie:
Vex only implements a 32-bit PA which does not take advantage of the potetnial 32-bit space in Sv32 mode. Very reasonably, Vex simply discards the top two unused bits.

However, the spec does require that the register occupy all 22 bits and it is possible for the OS to use the extra bits up top for some bookkeeping purpose. This commit proposes to expand the register to occupy the full 22 bits in case an OS is written to utilize the full width of the register as written in the spec.

I have tested the commit and confirmed it does not interfere with proper operation in Xous. I think it should also not cause any issues for Linux either, since it should just enable storing of an extra two bits an the top of the PPN that would normally be reported as zero.

### DIFF
--- a/src/main/scala/vexriscv/plugin/MmuPlugin.scala
+++ b/src/main/scala/vexriscv/plugin/MmuPlugin.scala
@@ -94,7 +94,8 @@ class MmuPlugin(ioRange : UInt => Bool,
       val satp = new Area {
         val mode = RegInit(False)
         val asid = Reg(Bits(9 bits))
-        val ppn = Reg(UInt(20 bits))
+        // Bottom 20 bits are used in implementation, but top 2 bits are still stored for OS use.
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi ^^

Ahhhh yes those 2 extra bits from SV32 to get 34 bits physical memory space XD

Sure, thanks for the PR :D

bunnie:
thanks for accepting the PR. I'm implementing a couple other kind of weird features that I'll send over as another PR from my other account (@buncram) related to the SATP.

Don't feel any pressure to accept them, mainly, I am implementing some security-oriented features that take advantage of the configurability of your core. I'm implementing them as options that are off-by-default so hopefully it's not a big deal to accept them into your main branch!

Dolu1990:
Ok :)

PR #289, opened 2022-12-21
Expose satp

buncram:
This PR adds a flag that allows the configuration to expose the SATP value for external inspection.

The primary motivation for this is to allow hardware to determine, when a system is running in virtual memory mode, if a given process is allowed to access sensitive data. The operative theory is that the SATP is intimately linked to the virtual process that is running, and by exposing this data outside the core we can make decisions about the state of the machine easily.

Hopefully the flag is coded in a way that users which do not care about this feature are not impacted by it. Let me know if you would prefer it coded in a different way for minimal impact on most users who do not share such security concerns.


Dolu1990:
Ahhh sorry i missed that PR.

I just pushed 2 changes to make things shorter.
Thanks for the PR :)

do not hesitate to ping me when there is no responses.

PR #294, opened 2023-01-12
Fix invalid hyperlink

chiangkd:
When I study about Briey Soc, I found there is a invalid hyperlink.
And I search the information from the official website


Dolu1990:
Thanks :)

PR #304, opened 2023-02-24
Fetcher: insert FORMAL_MODE encoded from privilegeService

lschuermann:
Previously, FORMAL_MODE would simply be hard-coded to "11", indicating machine mode. However, that's not necessarily true when using the CsrPlugin, which allows to switch the hart into either User or optional Supervisor mode. Hence we create a FORMAL_MODE insert in the fetch-phase (which is generally when the MPP register can take effect) and generate `rvfi_mode` based on that insert.


Dolu1990:
Thanks :D

PR #306, opened 2023-02-26
CsrPlugin: insert FORMAL_HALT := False

lschuermann:
The `FormalPlugin` requires a `FORMAL_HALT` insert to indicate whether the CPU will halt after retiring the current instruction. As this insert is provided only by the `HaltOnExceptionPlugin`, and the `CsrPlugin` also provides a conflicting `ExceptionService`, currently the `FormalPlugin` can't be used with the `CsrPlugin`. As there is no way to halt the CPU with the `CsrPlugin`, simply deassert `FORMAL_HALT`.


Dolu1990:
Thanks :D

PR #327, opened 2023-04-07
Remove sbt-assembly dependency

andreasWallner:
The plugin is not used in the VexRiscV build and causes issues for users since repo.scala-sbt.org seems to be down/sunset/?.

See also https://github.com/sbt/sbt/issues/7202

Updating the dependency would also have been an option, but since it's not used removal is easier.


Dolu1990:
Thanks ^^

Issue #347, opened 2023-06-09
Briey on AHBL


AdDraw:
Is it possible to change Briey's AXI crossbar to AHBL easily?

 I have tried a couple of times but all I get are errors.
 
 I have put the code on https://github.com/AdDraw/VexRiscv/blob/briey_on_ahbl/src/main/scala/vexriscv/demo/Briey.scala.
 
 
 Errors are mostly related COMB loops at the ram <> dBus connection as far as I can tell.
 > Error log is in https://github.com/SpinalHDL/VexRiscv/commit/99cf310de79b0200995535b073ac518380be324b
 
 I know that AXI crossbar factor can add Pipelining between different AXI channels and that this option is not available for AHBLite crossbars.
 
 How can I remove those COMB loops?

Dolu1990:
Hi,

You can try to set both : 
- dBusCmdSlavePipe
- dBusRspSlavePipe
in the DBusCachedPlugin, this will isolate it.

https://github.com/SpinalHDL/VexRiscv/blob/c52433575dec04f10063b2fd7cebd0545c8b1be9/src/main/scala/vexriscv/plugin/DBusCachedPlugin.scala#L52

Should goes better then i think ^^

AdDraw:
As far as I can tell, `DBusCachedPlugin`'s dBus of type `DataCacheMemBus` does not support `toAxiLite3Master()` or any other `AxiLite3` type conversions.

Dolu1990:
Hooo ok ^^
Then you will need to add those pipelines in some other ways
Murax kind of do that : 
https://github.com/SpinalHDL/VexRiscv/blob/c52433575dec04f10063b2fd7cebd0545c8b1be9/doc/gcdPeripheral/src/main/scala/vexriscv/demo/Murax.scala#L266

AdDraw:
I have applied this to both IBus and DBus I can compile it now, thanks

What would be the best way to test that it still works? `src/test/cpp/briey`?

Dolu1990:
So, normaly, you can avoid having it on IBus, in order to preserve the performance.

> What would be the best way to test that it still works? src/test/cpp/briey?

Hmm there is nothing automated for the SoC level testing. So, i guess as long as it can run some software (ex dhrystone) and doesn't catch fire that's ok.

> I can close this now, @Dolu1990 would there be any merit in opening a PR for this?

I think yes, a function could be added to DBusSimpleBus, which would do your : 

```scala
[...]

PR #350, opened 2023-06-15
Add cmd halfPipe function to DBusSimpleBus

AdDraw:
Motivated by https://github.com/SpinalHDL/VexRiscv/issues/347#issuecomment-1587520015.

Adds a function called `cmdHalfPipe` to `DBusSimpleBus` which allows for a simple `halfPipe()` insertion on the `cmd` Stream in order to break combinatorial loops(when necessary).

Tested manually on board with Briey that used AhbLite3 interconnect(which forces the use of DBusSimplePlugin since Cached plugin does not have the `toAhbLite3Master()` function)

### DIFF
--- a/src/main/scala/vexriscv/plugin/DBusSimplePlugin.scala
+++ b/src/main/scala/vexriscv/plugin/DBusSimplePlugin.scala
@@ -103,6 +103,13 @@ case class DBusSimpleBus(bigEndian : Boolean = false) extends Bundle with IMaste
     s
   }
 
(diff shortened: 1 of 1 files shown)


Dolu1990:
Thanks ^^

Dolu1990:
Ahh my bad, just pushed a fix to my change : 
https://github.com/SpinalHDL/VexRiscv/commit/5860dc2321a9b41737fb094270c067b8acede46c

PR #377, opened 2023-11-13
Fix ambiguous function call to bind()

Tectu:
The call to `bind()` can actually resolve to `std::bind()` instead of libc's `bind()`.
This fix ensures that we're definitely always calling the correct `bind()`.

Took me way too long to hunt this one down...

This happened with `clang++ 16.0.6`.


Dolu1990:
Damned XD

Thanks :D

PR #386, opened 2023-12-21
Add mill to compile and test VexRiscv

davine47:
Hi, sbt is a very heavy tool to build scala files, instead, mill is very simple and benefit to use in offline environment.
So i add whole mill compile and test scripts.


Dolu1990:
Thanks :)

PR #388, opened 2024-01-31
Perf counters

### LINKED ISSUE #361: RiscV Counters
Since I need them for a project, I was wondering, if the performance counters will be implemented.
I might try building them myself, but I feel not as familiar with either Spinal or VexRiscv yet.

jjjt-git:
Fixes #361
Not tested, because I did not find the time for it.
So I do not know, whether it works or not.

### DIFF
--- a/src/main/scala/vexriscv/demo/VexRiscvAhbLite3.scala
+++ b/src/main/scala/vexriscv/demo/VexRiscvAhbLite3.scala
@@ -168,7 +168,7 @@ object VexRiscvAhbLite3{
 //            // On Artix FPGA jtag :
 //            val jtagCtrl = JtagTapInstructionCtrl()
 //            val tap = jtagCtrl.fromXilinxBscane2(userId = 1)
-//            jtagCtrl <> plugin.io.bus.fromJtagInstructionCtrl(ClockDomain(tap.TCK))
+//            jtagCtrl <> plugin.io.bus.fromJtagInstructionCtrl(ClockDomain(tap.TCK),0)
           }
           case _ =>
         }
--- a/src/main/scala/vexriscv/demo/VexRiscvAhbLite3.scala
+++ b/src/main/scala/vexriscv/demo/VexRiscvAhbLite3.scala
@@ -122,8 +122,12 @@ object VexRiscvAhbLite3{
               mscratchGen    = false,
               mcauseAccess   = CsrAccess.READ_ONLY,
               mbadaddrAccess = CsrAccess.READ_ONLY,
+              mcycleAccess   = CsrAccess.NONE,
+              minstretAccess = CsrAccess.NONE,
               ecallGen       = false,
-              wfiGenAsWait         = false
+              wfiGenAsWait         = false,
+              ucycleAccess   = CsrAccess.NONE,
+              uinstretAccess = CsrAccess.NONE
             )
           ),
           new YamlPlugin("cpu0.yaml")
--- a/src/main/scala/vexriscv/demo/VexRiscvAhbLite3.scala
+++ b/src/main/scala/vexriscv/demo/VexRiscvAhbLite3.scala
@@ -122,12 +122,8 @@ object VexRiscvAhbLite3{
               mscratchGen    = false,
               mcauseAccess   = CsrAccess.READ_ONLY,
               mbadaddrAccess = CsrAccess.READ_ONLY,
-              mcycleAccess   = CsrAccess.NONE,
-              minstretAccess = CsrAccess.NONE,
               ecallGen       = false,
-              wfiGenAsWait         = false,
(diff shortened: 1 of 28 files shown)


Dolu1990:
Ok thanks :)

jjjt-git:
The Plugin works.

src/main/scala/vexriscv/demo/VexRiscvAhbLite3.scala
```diff
@@ -122,12 +122,8 @@ object VexRiscvAhbLite3{
               mscratchGen    = false,
               mcauseAccess   = CsrAccess.READ_ONLY,
               mbadaddrAccess = CsrAccess.READ_ONLY,
-              mcycleAccess   = CsrAccess.NONE,
```
Dolu1990, line 125: Would it be possible to still keep those option ?
I know that would be redundant to the CounterPlugin implementation, but that would preserve backward compatibility without any disruption. (not break users setups)

Dolu1990: (this is a general comment, so would also apply for the other demo)

src/main/scala/vexriscv/demo/smp/VexRiscvSmpCluster.scala
```diff
@@ -150,7 +150,7 @@ class VexRiscvSmpClusterWithPeripherals(p : VexRiscvSmpClusterParameter) extends
     plic.addTarget(core.cpu.externalSupervisorInterrupt)
     List(clint.logic, core.cpu.logic).produce {
       for (plugin <- core.cpu.config.plugins) plugin match {
-        case plugin: CsrPlugin if plugin.utime != null => plugin.utime := clint.logic.io.time
```
Dolu1990, line 153: Same here, would need to preserve that feature in the CsrPlugin. (even if that is a duplication) to not break people code

src/main/scala/vexriscv/demo/VexRiscvAhbLite3.scala
jjjt-git: The options are still available in the CSRPlugin. I removed them from here and other places, since CsrAccess.NONE was and is the default value for these.

src/main/scala/vexriscv/demo/smp/VexRiscvSmpCluster.scala
jjjt-git: The feature is still in CSRPlugin, even if disabled when CounterService is present (which in this demo it is), so it should break no existing code which uses the Counters in CsrPlugin.

src/main/scala/vexriscv/demo/VexRiscvAhbLite3.scala
Dolu1990: Would be possible to revert all the removals of "mcycleAccess = CsrAccess.NONE" and similar things ? 

They are there to make it clear to the user that it isn't included, and that if they want they can just turn it on there.

jjjt-git:
I reverted the removal of the config options.
The GenFull still uses the CounterPlugin, as does the VexRiscvAxi4LinuxPlicClint.
However there is (currently) no module in the repo to provide performance counters. It should be quite simple to add to the various cache-modules, but I have not looked into it yet.

Dolu1990:
Thanks ^^
The pr seems all good to me, i just updated it to go into dev instead of master.

Also note since november i'm working on VexiiRiscv : 
https://github.com/SpinalHDL/VexiiRiscv

medium / long terms the idea is to migrate toward it.
Let's me know if you are interrested into the project ^^

src/main/scala/vexriscv/plugin/CsrPlugin.scala
```diff
@@ -649,7 +654,7 @@ class CsrPlugin(val config: CsrPluginConfig) extends Plugin[VexRiscv] with Excep
     pipeline.update(MPP, UInt(2 bits))
 
     if(withExternalMhartid) externalMhartId = in UInt(mhartidWidth bits)
-    if(utimeAccess != CsrAccess.NONE) utime = in UInt(64 bits) setName("utime")
+    if(pipeline.serviceExist(classOf[CounterService]) && utimeAccess != CsrAccess.NONE) utime = in UInt(64 bits) setName("utime")
```
Dolu1990, line 657: Ahhh this seems broken, shouldn't it be  !pipeline.serviceExist(classOf[CounterService]) ?

jjjt-git: Yes, this must have slipped by.

jjjt-git:
Fixed the condition.

Dolu1990:
Thanks ^^

jjjt-git:
my pleasure

PR #397, opened 2024-03-10
Fix SMP compile-time error when disabling supervisor option

cherrypiejam:
When generating SMP configuration with supervisor disable, the compiler stucks at waiting for the signal from
`externalSupervisorInterrupt`, which is generated conditionally based on `withSupervisor` in CsrPlugin.


Dolu1990:
Thanks ^^

PR #399, opened 2024-03-27
Handle `ERR` in `toWishbone`

martijnbastiaan:
This PR implements handling `ERR` in `toWishbone`. 

Note that the following may look strange:

```scala
  cmdBridge.ready := cmdBridge.valid && (bus.ACK || bus.ERR)
```

But this is imposed by the Wishbone spec:

> *Rule 3.45*
> If a SLAVE supports the `[ERR_O]` or `[RTY_O]` signals, then the SLAVE MUST NOT assert more than one of the following signals at any time: `[ACK_O]`, `[ERR_O]` or `[RTY_O]`.

----------

Questions for reviewers:

 * Is there an example regression test for, say, AXI error handling I could use as a base for a Wishbone test?
 * Where can I add a changelog entry (if applicable)?

### DIFF
--- a/src/main/scala/vexriscv/ip/DataCache.scala
+++ b/src/main/scala/vexriscv/ip/DataCache.scala
@@ -368,13 +368,13 @@ case class DataCacheMemBus(p : DataCacheConfig) extends Bundle with IMasterSlave
     bus.WE  := cmdBridge.wr
     bus.DAT_MOSI := cmdBridge.data
 
-    cmdBridge.ready := cmdBridge.valid && bus.ACK
+    cmdBridge.ready := cmdBridge.valid && (bus.ACK || bus.ERR)
     bus.CYC := cmdBridge.valid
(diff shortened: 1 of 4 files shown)


Dolu1990:
Hi,

Thanks for the fix ^^

> Is there an example regression test for, say, AXI error handling I could use as a base for a Wishbone test?

So the issue is that those bridges are more used inside SoC, and tested at a SoC level. There is no proper test for it in VexRIscv repo. Not great.

> Where can I add a changelog entry (if applicable)?

So far, the git commits are the logs

Note that i'm now working on VexiiRiscv : https://github.com/SpinalHDL/VexiiRiscv
Aiming for the mooooooon XD

Cheers
Charles

martijnbastiaan:
Thanks for the merge!

> Note that i'm now working on VexiiRiscv : https://github.com/SpinalHDL/VexiiRiscv

Good luck!!

PR #410, opened 2024-06-04
Exposed write mask on default iBus

### LINKED ISSUE #409: default bus doesn't expose write mask
Write mask is not visible outside of the core when using default bus

MrJake222:
Fixes #409


Dolu1990:
Thanks ^^

PR #413, opened 2024-06-17
Fix Mhz -> MHz in README, comments and Dhrystone benchmark output

mrcmry:
SI units are case sensitive and analog engineers are very sensitive to this :)

It seems that the Dhrystone benchmark is done with .asm files, perhaps generated from [here](https://github.com/Keith-S-Thompson/dhrystone), where [Hz](https://en.wikipedia.org/wiki/Hertz) is wrongly spelled "hz".

This PR fix this in the README, the scala code that seems to generate the README but not the .asm files.

Ideally a fix in the upstream C code repos would also be  needed, but IMHO it's not a good idea, because it would break too many tests like this one.

Note that I didn't test the fix in the scala file because I don't have a running VexRiscv setup.


Dolu1990:
Thanks ^^

PR #423, opened 2024-08-27
Tunneled EmbeddedRiscvJtag without TAP

craigjb:
This change allows tunneling JTAG (e.g. with Xilinx BSCANE2) to the VexRiscv EmbeddedRiscvJtag plugin. The plugin required another case to handle tunneling without creating a TAP.

This change also adds a section to README.md that shows how to use it with Xilinx BSCANE2.

I successfully tested it using a "smallest" configuration on a XC7A200T-FBG676 with a Digilent HS3 adapter and upstream riscv-openocd.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1404,6 +1404,62 @@ halt
 
 A full example can be found in GenFullWithOfficialRiscvDebug.scala
 
+##### Tunneled JTAG
+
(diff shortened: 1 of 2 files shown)


Dolu1990:
Thanks :D

PR #427, opened 2024-09-04
DBusSimplePlugin: don't force SEL to 1111 on read.

### LINKED ISSUE #426: DBusSimplePlugin's Wishbone support forces 32-bit reads
I notice that `DBusSimplePlugin`'s Wishbone bridge forces all reads to appear to be 32-bits wide:

https://github.com/SpinalHDL/VexRiscv/blob/0af9894e698288d65f6127c4a2806986dbabf09c/src/main/scala/vexriscv/plugin/DBusSimplePlugin.scala#L197-L199

I'm not sure of the exact reason it was done this way, but I'm wondering if you'd be amenable to changing this default! For some MMIO it's helpful to be able to distinguish the read size.

kivikakk:
Closes #426.

Thanks! :)


Dolu1990:
thanks ^^

kivikakk:
Thanks! 🐱

PR #434, opened 2024-11-13
add Murax config with native jtag based on the docs

goekce:
I will use the native Jtag on AMD FPGAs often, so I created a Murax config according to the docs in:

https://github.com/SpinalHDL/VexRiscv/tree/master/doc/nativeJtag

It was my first experience with Scala. I would be glad to hear your feedback on whether my approach with `ifGen` is plausible. I had doubts about whether I should create an independent `class` or not.

### DIFF
--- a/doc/nativeJtag/README.md
+++ b/doc/nativeJtag/README.md
@@ -24,47 +24,15 @@ reader is capable of generating the Murax SoC as it is described there.
 The BSCANE2 allows access between the internal FPGA logic and the JTAG Boundary Scan logic controller. This allows
 for communication between the internally running design and the dedicated JTAG pins of the FPGA.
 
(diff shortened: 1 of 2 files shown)


goekce:
> That should be fine.

You probably refer to my approach with `ifGen`, no?

I indeed could not find any `jtagNative` string in the generated code `Murax.v`. Now I moved the names outside and they are visible.

I believe they don't belong to val `io` because `io` is only for generating the interface signals, right?

Is this what you recommended?

Dolu1990:
> You probably refer to my approach with ifGen, no?

Ahh the whole PR ^^ 

> I believe they don't belong to val io because io is only for generating the interface signals, right?

Yes right, io is just for the component in/out

Look all good.
Thanks :D

PR #442, opened 2025-02-01
Add address granularity for wishbone

jdavidberger:
Add the wishbone granularity attribute to the conversion functions for iBus / dBus


Dolu1990:
Thanks ^^

PR #444, opened 2025-02-14
Make TDATA2 RW. Openocd requires this for hw breakpoints

jdavidberger:
https://riscv.org/wp-content/uploads/2024/12/riscv-debug-release.pdf

> Most trigger functionality is optional. All tdata registers follow write-any-read-legal semantics. If a
debugger writes an unsupported configuration, the register will read back a value that is supported
(which may simply be a disabled trigger). This means that a debugger must always read back
values it writes to tdata registers, unless it already knows already what is supported. Writes to
one tdata register may not modify the contents of other tdata registers, nor the configuration of
any trigger besides the one that is currently selected.

OpenOCD in particular reads back this value and if it doesn't match; won't use the breakpoint


Dolu1990:
Thanks ^^

PR #447, opened 2025-04-16
Bump to Spinal version 1.12.0

holopoggers:
Codebase requires int support in isPow2, the latest version of spinal has this support.


Dolu1990:
Thanks ^^

PR #450, opened 2025-05-24
Fix DBUS_SIMPLE_AHBLITE3 dBusAccess param error

SofanHe:
# Description

Fixed the issue of incorrect parameters in the dBusAccess function call in class 'DBusSimpleAhbLite3'.

The updated version now supports regression test under the 'VexRiscvAhbLite3' configuration.


Dolu1990:
Thanks :D

PR #454, opened 2025-07-18
EmbeddedRiscvJtag support for system bus

craigjb:
# Description
This change adds an optional system bus sysBus to the EmbeddedRiscvJtag plugin to support 32-bit read+write accesses per RISC-V Debug Specification v1.0.

When enabled using `withSysBus = true`, the plugin exposes a `DBusSimpleBus` connected to the DebugModule's system bus. Only 32-bit addresses and accesses are supported for now, but this change could be extended for 128, 64, 16, or 8 bit accesses (see referenced SpinalHDL PR)

**Depends on this SpinalHDL PR**
https://github.com/SpinalHDL/SpinalHDL/pull/1768
(DebugModule system bus support)

# Motivation and Context
Without system bus support, debug memory accesses must go through the CPU. I noticed this when building Rust firmware for a VexRiscV based system with RTT logging. When running the firmware with probe-rs, the host continuously looks for new logs in a ring buffer in memory. Without system bus support, this causes delays in the firmware's execution during those memory polls--not helpful when you're working on something with real-time requirements!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1465,6 +1465,41 @@ init
 halt
 ```
 
+##### System Bus
+The EmbeddedRiscvJtag plugin can also be configured to support system bus accesses per the RISC-V Debug Specification v1.0. This allows memory access without impacting CPU execution and can be handy when a debug host is regularly polling (e.g. RTT logging).
+
+```scala
+new EmbeddedRiscvJtag(
+  p = DebugTransportModuleParameter(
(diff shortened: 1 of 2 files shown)


vojtech-arcsec:
Hi! Memory access without halting the core would indeed be very useful. Are there plans for merging this PR or implementing it in some other way? Thanks!

PR #468, opened 2025-12-14
Add mstatush register

### LINKED ISSUE #462: Compatibility with OpenSBI 1.6 or newer
Hello there,

Foremost, many thanks for this excellent project, I have been using the RISC-V core for almost one year now in different designs.

Recently, I tried to boot with and OpenSBI version equivalent or newer than version 1.6; however, it never went through OpenSBI to the next bootloader. After some investigation, I've noticed that OpenSBI tried to access one CSR that is not yet available in VexRiscv but which has been apparently marked mandatory in the ratified RISC-V 1.11 specification. The missing CSR is `MSTATUSH` (`0x310`) which is an additional machine status register.

If there are plans to support this CSR at some point, I would happily help to contribute to that milestone.

Thanks

rvodden:
Fixed #462

### DIFF
--- a/src/test/cpp/raw/mstatush/makefile
+++ b/src/test/cpp/raw/mstatush/makefile
@@ -1,0 +1,3 @@
+PROJ_NAME=mstatush
+
+include ../common/asm.mk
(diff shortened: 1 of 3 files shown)


Dolu1990:
Nice thanks ^^

PR #499, opened 2026-09-19
Follow the SWD library split in SpinalHDL

disdi:
SpinalHDL moved the SWD wire side (Swd bundle, SwdPhy, SwdDp) into spinal.lib.com.swd, and DebugTransportModuleSwd.io.swd is now slave(Swd()) with swdio as a TriState.

The SMP cluster maps swdio.read/write/writeEnable onto the same pinned top-level names as before (debugPort_swdio_i/o/oe), so the LiteX black-box port map is unchanged and the generated netlist is identical up to internal port renames.

The SWD PHY, DP and DTM suites and their sim driver never depended on VexRiscv; they now live in SpinalHDL's tester next to the code they test.

Requires the matching  [SpinalHDL/1966](https://github.com/SpinalHDL/SpinalHDL/pull/1966).


Dolu1990:
Thanks ^^

